PR #5, opened 2015-06-10
F/shims for ie8

alavers:


alavers:
Note that this change only puts the javascript in working order. The HTML+CSS render quite badly, but it's functional.

Also, in IE8 the appUserId is persisted, but any existing conversation messages fail to get fetched on page refresh. To be fixed later.

src/js/main.js
```diff
@@ -2,12 +2,14 @@
 
 'use strict';
 
+require('./compat');
```
juliangarritano: Maybe drop these in bootstrap instead?

example/template/demo.html
```diff
@@ -6,7 +6,7 @@
     <meta charset="utf-8">
     <title>Conversation Demo</title>
     <script src="http://code.jquery.com/jquery-1.11.3.min.js"></script>
-    <script src="/dist/supportkit.js"></script>
+    <script src="/dist/WIDGET_CODE"></script>
```
juliangarritano: Cool.

juliangarritano:
:+1:

src/js/main.js
lemieux: +1

lemieux:
:+1:

PR #18, opened 2015-06-25
Replace Modernizr by a simpler check.

lemieux:
Modernizr seems to break for some users. Let's replace it by something simpler.


alavers:
:+1:

PR #25, opened 2015-06-25
Track API

lemieux:
I refactored the boot part to avoid creating a conversation every time and I added the track API for the sdk.

### DIFF
--- a/src/js/controllers/chatController.js
+++ b/src/js/controllers/chatController.js
@@ -2,9 +2,9 @@
 
 var $ = require('jquery'),
     cookie = require('cookie'),
-    bindAll = require('lodash.bindall');
-
-var ViewController = require('view-controller');
+    bindAll = require('lodash.bindall'),
+    _ = require('underscore'),
+    ViewController = require('view-controller');
 
 var endpoint = require('../endpoint'),
     vent = require('../vent'),
@@ -14,7 +14,7 @@ var Conversation = require('../models/conversation');
 
 var ChatView = require('../views/chatView'),
     HeaderView = require('../views/headerView'),
-    ConversationView =  require('../views/conversationView');
+    ConversationView = require('../views/conversationView');
 
 var ChatInputController = require('../controllers/chatInputController');
 
@@ -30,6 +30,8 @@ module.exports = ViewController.extend({
         this.isOpened = false;
         this.user = this.getOption('user');
         this.uiText = this.getOption('uiText') || {};
+
+        this.conversationInitiated = false;
 
         this.listenTo(this.user, 'change:conversationStarted', this.onConversationStarted);
     },
@@ -142,9 +144,21 @@ module.exports = ViewController.extend({
     },
 
     _initConversation: function() {
-        return this.collection.fetch()
-            .then(this._getConversation)
-            .then(this._initFaye);
+        var promise;
+
+        if (this.conversationInitiated) {
+            promise = this._getConversation();
+        } else {
+            promise = this.collection.fetch()
+                .then(this._getConversation)
+                .then(this._initFaye)
+                .then(_.bind(function(conversation) {
+                    this.conversationInitiated = true;
+                    return conversation;
+                }, this));
+        }
+
+        return promise;
     },
 
     _initMessagingBus: function(conversation) {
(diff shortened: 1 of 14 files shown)


mspensieri:
Awesome :+1:

src/js/controllers/chatController.js
```diff
@@ -10,9 +10,11 @@ var endpoint = require('../endpoint'),
     vent = require('../vent'),
     faye = require('../faye');
 
+var Conversation = require('../models/conversation');
+
 var ChatView = require('../views/chatView'),
     HeaderView = require('../views/headerView'),
-    ConversationView = require('../views/conversationView');
+    ConversationView =  require('../views/conversationView');
```
alavers: jsfmt disabled? :)

lemieux: lol... It was my test for atom-jsfmt and apparently it failed it...

src/js/controllers/chatController.js
```diff
@@ -58,12 +60,30 @@ module.exports = ViewController.extend({
...
+        var conversationDeferred = $.Deferred(),
+            messageDeferred = $.Deferred();
+
+        if (this.conversation.isNew()) {
+            this.conversation = this.collection.create(this.conversation, {
+                success: conversationDeferred.resolve,
+                error: conversationDeferred.reject
+            });
+
+            conversationDeferred.then(this._initFaye);
```
alavers: Should we rely on a `change:conversationStarted` to call `initFaye` for us?

src/js/main.js
```diff
@@ -53,6 +60,10 @@ var SupportKit = Marionette.Object.extend({
     },
 
     init: function(options) {
+        if (this.ready) {
```
alavers: An important bit of code for @juliangarritano to keep in mind for app uesr login/logout

src/js/controllers/chatController.js
lemieux: Yeah I think we should.

src/js/main.js
```diff
@@ -108,8 +115,34 @@ var SupportKit = Marionette.Object.extend({
...
+                    url: urljoin('appusers', res.appUserId, 'event'),
+                    model: Event
+                });
+
+                var RuleCollection = Backbone.Collection.extend({
+                    model: Rule
+                });
+
+                // this._events overrides some internals for event bindings in Backbone
+                this._eventCollection = new EventCollection(res.events, {
```
alavers: Looking ahead to the app user story, it might be easier to track this inside of a user object or controller, unless its OK to simply re-write this var with every init() call.

alavers: In general the either main.js and our controllers need to tolerate all user data (conversation, events, rules) getting swapped out at any time,

or,

every call to init() simply nukes and resets all model+controller state

src/js/controllers/chatController.js
lemieux: While trying this, I think we don't want to throttle `updateUser` but debounce it with a maxWait. Currently, if you call updateUser and you are being throttle, you lose that information unless you reupdate the user outside of the throttle window.

src/js/main.js
lemieux: By swapped out, you mean resetted?

There's a `destroy` method now on the SupportKit object that nukes everything (collection, models, events, and the widget itself) and it can be re-init after that. Why would you re-init SupportKit without destroying it first?

src/js/main.js
```diff
@@ -172,6 +205,58 @@ var SupportKit = Marionette.Object.extend({
...
+
+        if (rulesContainEvent) {
+            this._eventCollection.create({
+                name: eventName,
+                user: this.user
+            }, {
+                success: _.bind(this._checkConversationState, this)
+            });
+        } else if (!rulesContainEvent && !hasEvent) {
+            this._createEvent(eventName);
```
alavers: Could maybe merge `_createEvent` and `_hasEvent` into a single `_ensureEventExists()`

src/js/main.js
```diff
@@ -186,8 +271,20 @@ var SupportKit = Marionette.Object.extend({
         }, this));
 
     },
+
     onReady: function() {
         this.ready = true;
+        this.track('skt-appboot');
+    },
+
+    onDestroy: function() {
```
juliangarritano: Beautiful.

src/js/main.js
alavers: Made this comment before seeing the destroy method :) Yes would definitely use that in the switch user case.

lemieux: :)

alavers:
Rest looks good :+1:

src/js/main.js
lemieux: That'd be harsh on the unit tests and make that async. I could create `_ensureEventExists` and call `_hasEvent` in there.

PR #26, opened 2015-06-26
F/app user login

juliangarritano:
Added the ability to log in / log out based on given userId paired with devideId.
Added option to include jwtToken in init call for future ability to securely log in users.
Added ability to log out (clearing deviceId from cookies).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ SupportKit.init({
     surname: 'Person',
     email: 'their_email@whatever.com',
     // For secure mode
-    jwtToken: 'your_jwt_token',
+    jwt: 'your_jwt',
     userId: 'user_id',
     // Additional properties
     properties: {
--- a/README.md
+++ b/README.md
@@ -61,6 +61,9 @@ SupportKit.init({
     givenName: 'Cool',
     surname: 'Person',
     email: 'their_email@whatever.com',
+    // For secure mode
+    jwtToken: 'your_jwt_token',
+    userId: 'user_id',
     // Additional properties
     properties: {
         'anything': 'whatever_you_want'    
--- a/example/template/demo.html
+++ b/example/template/demo.html
@@ -26,6 +26,6 @@
 </head>
   <body>
     <h1>Test</h1>
-    <button onclick="SupportKit.logout(); console.log('User succesfully logged out.')">Log Out</button>
+    <button onclick="SupportKit.logout(); document.location.reload(true);">Log Out</button>
   </body>
 </html>
--- a/example/template/demo.html
+++ b/example/template/demo.html
@@ -14,13 +14,18 @@
             appToken: 'APP_TOKEN',
             givenName: 'GIVEN_NAME',
             surname: 'SURNAME',
+            email: 'EMAIL',
+            jwtToken: 'JWT_TOKEN',
+            userId: 'USER_ID',
             properties: {
               'TEST': true
             }
         });
     </script>
+
 </head>
   <body>
     <h1>Test</h1>
+    <button onclick="SupportKit.logout(); console.log('User succesfully logged out.')">Log Out</button>
   </body>
 </html>
--- a/example/template/demo.html
+++ b/example/template/demo.html
@@ -15,7 +15,7 @@
             givenName: 'GIVEN_NAME',
             surname: 'SURNAME',
             email: 'EMAIL',
-            jwtToken: 'JWT_TOKEN',
+            jwt: 'JWT',
             userId: 'USER_ID',
             properties: {
               'TEST': true
--- a/src/js/endpoint.js
+++ b/src/js/endpoint.js
@@ -6,18 +6,21 @@ module.exports.rootUrl = ROOT_URL;
 
 // State params set by main
 module.exports.appToken = undefined;
-module.exports.jwtToken = undefined;
+module.exports.jwt = undefined;
 module.exports.appUserId = undefined;
 
 module.exports._rest = function(method, path, body) {
     var deferred = $.Deferred();
+    var headers = {
+        'app-token': this.appToken
+    };
+
+    this.jwt ? headers.Authorization = 'Bearer ' + this.jwt : '';
+
     $.ajax({
         url: urljoin(this.rootUrl, path),
         type: method,
-        headers: {
-            'app-token': this.appToken,
-            'Authorization': 'Bearer ' + this.jwtToken
-        },
+        headers: headers,
         data: JSON.stringify(body),
         contentType: 'application/json',
         success: function(res) {
--- a/src/js/endpoint.js
+++ b/src/js/endpoint.js
@@ -50,3 +50,9 @@ module.exports.post = function(path, body) {
 module.exports.put = function(path, body) {
     return this._rest('PUT', path, body);
 };
+
+module.exports.reset = function() {
+    delete this.appToken;
+    delete this.jwt;
+    delete this.appUserId;
+};
--- a/src/js/endpoint.js
+++ b/src/js/endpoint.js
@@ -6,6 +6,7 @@ module.exports.rootUrl = ROOT_URL;
 
 // State params set by main
 module.exports.appToken = undefined;
+module.exports.jwtToken = undefined;
 module.exports.appUserId = undefined;
 
 module.exports._rest = function(method, path, body) {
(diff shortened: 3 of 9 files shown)


src/js/main.js
```diff
@@ -100,7 +98,7 @@ var SupportKit = Marionette.Object.extend({
                 browserLanguage: navigator.language,
                 currentUrl: document.location.href,
                 sdkVersion: this.VERSION,
-                currentTitle: document.title
+                currentTitle: document.title,
```
juliangarritano: Oops. Will fix.

test/specs/main.spec.js
```diff
@@ -30,25 +33,65 @@ describe('Main', function() {
         // those tests are using the expect form since undefined
         // cannot be tested with the should syntax
         it('should publish a global', function() {
-            expect(global.SupportKit).to.exist;
+            global.SupportKit.should.exist;
         });
 
-        it('should not publish dependencies in global context', function() {
+        it('should not publish dep`encies in global context', function() {
```
juliangarritano: Will fix.

juliangarritano:
Closed until ready.

README.md
```diff
@@ -61,6 +61,9 @@ SupportKit.init({
     givenName: 'Cool',
     surname: 'Person',
     email: 'their_email@whatever.com',
+    // For secure mode
+    jwtToken: 'your_jwt_token',
```
alavers: nit, Token is redundant. Like ATM Machine

src/js/endpoint.js
```diff
@@ -14,14 +15,15 @@ module.exports._rest = function(method, path, body) {
         url: urljoin(this.rootUrl, path),
         type: method,
         headers: {
-            'app-token': this.appToken
+            'app-token': this.appToken,
+            'Authorization': 'Bearer ' + this.jwtToken
```
alavers: jwt is optional, so this header shouldn't be there at all if its unspecified

src/js/main.js
```diff
@@ -122,6 +120,30 @@ var SupportKit = Marionette.Object.extend({
             .done();
     },
 
+    logout: function() {
+        // delete the deviceid cookie
+        document.cookie = SK_STORAGE + '=; expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+    },
+
+    checkLocalStorage: function(userId) {
```
alavers: Is this method only used by getDeviceId? If so, maybe it doesn't deserve to be its own one liner method

alavers:
Rest looks good :+1:

example/template/demo.html
```diff
@@ -14,13 +14,18 @@
             appToken: 'APP_TOKEN',
             givenName: 'GIVEN_NAME',
             surname: 'SURNAME',
+            email: 'EMAIL',
```
lemieux: not sure how we should handle this page. I keep comment portion of it to test different scenarios.

src/js/endpoint.js
```diff
@@ -6,22 +6,27 @@ module.exports.rootUrl = ROOT_URL;
...
+module.exports.jwt = undefined;
 module.exports.appUserId = undefined;
 
 module.exports._rest = function(method, path, body) {
     var deferred = $.Deferred();
+    var headers = {
+        'app-token': this.appToken
+    };
+
+    this.jwt ? headers.Authorization = 'Bearer ' + this.jwt : '';
```
lemieux: use if/else instead

src/js/main.js
```diff
@@ -122,6 +120,26 @@ var SupportKit = Marionette.Object.extend({
...
+    logout: function() {
+        // delete the deviceid cookie
+        document.cookie = SK_STORAGE + '=; expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+    },
+
+    getDeviceId: function(options) {
+        var userId = options.userId, deviceId;
+
+        // get device ID first from local storage, then cookie. Otherwise generate new one
+        deviceId = userId && localStorage.getItem(SK_STORAGE + '_' + userId);
```
lemieux: you could simplify this : 

```
 deviceId = userId && localStorage.getItem(SK_STORAGE + '_' + userId)
                   || cookie.parse(document.cookie)[SK_STORAGE]
                   || uuid.v4().replace(/-/g, '');
```

src/js/mixins.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
         options.beforeSend = function(xhr) {
             xhr.setRequestHeader('app-token', endpoint.appToken);
             xhr.setRequestHeader('Content-Type', 'application/json');
+            xhr.setRequestHeader('Authorization', 'Bearer ' + endpoint.jwt);
```
lemieux: don't add the header if no jwt

test/specs/main.spec.js
```diff
@@ -40,15 +43,55 @@ describe('Main', function() {
     });
 
     describe('#init', function() {
+        var userId = 'thisisauserid',
+            appToken = 'thisisanapptoken',
+            jwt = 'thisisajwt';
```
lemieux: can we do a second test instead with the token. We need to test both flows.

src/js/main.js
```diff
@@ -122,6 +120,26 @@ var SupportKit = Marionette.Object.extend({
             .done();
     },
 
+    logout: function() {
+        // delete the deviceid cookie
+        document.cookie = SK_STORAGE + '=; expires=Thu, 01 Jan 1970 00:00:01 GMT;';
```
lemieux: Hmm.. this is nice, but how does it logs out the user? Unless `destroy` is called before that, the widget is still fully operational. Collections are still initialized and the headers are set up to use the previous id. Also, is the JWT per user or per app? If per user, it needs to be cleared too. 

Shouldn't we just bundle or alias `logout` and `destroy` and add the cookie clearing logic in the `onDestroy` handler?

test/specs/main.spec.js
```diff
@@ -73,15 +116,15 @@ describe('Main', function() {
         });
 
         it('should call _updateUser', function() {
-            expect(SupportKit._updateUser).to.be.calledOnce;
```
lemieux: thank you :)

src/js/endpoint.js
juliangarritano: Yeah whenever I'm in a situation where I want to do something like that it takes me like 5 minutes to decide one way or the other.
I wish there was an operator that allowed it.

src/js/main.js
juliangarritano: Absolutely.
At the time I wrote this destroy didn't exist.

src/js/endpoint.js
lemieux: well ... this would be better : `headers.Authorization = this.jwt? 'Bearer ' + this.jwt : ''`

lemieux: I was just surprised to see an assignment in there instead of being the logic to select the good value for the assignment.

src/js/main.js
juliangarritano: JWT should be per app... But I think that we should still clear it onDestroy.

lemieux: I agree. To make the widget work again, `init` must be called again anyway.

example/template/demo.html
juliangarritano: Setting the strings to empty is insufficient?

lemieux: Maybe.. I actually never tried... :/

src/js/endpoint.js
juliangarritano: Right... But I don't want the alternative you mentioned because that still adds the key to the Object.

test/specs/main.spec.js
juliangarritano: ?

src/js/endpoint.js
lemieux: True! if/else it is then!

example/template/demo.html
juliangarritano: In the config file! Not this one. But yeah, try that.

test/specs/main.spec.js
lemieux: With and without the token? This test will only test the case where we have a token.

src/js/main.js
```diff
@@ -155,6 +151,25 @@ var SupportKit = Marionette.Object.extend({
...
+        // reset the cookie and local storage
+        document.cookie = SK_STORAGE + '=' + deviceId;
+        userId && localStorage.setItem(SK_STORAGE + '_' + userId, deviceId);
+
+        return deviceId;
+    },
+
     resetUnread: function() {
         this._checkReady();
         this._chatController._resetUnread();
```
lemieux: I guess `_resetUnread` in the chatController should be `resetUnread` if it's called from here.

lemieux:
:+1:

PR #28, opened 2015-07-02
Fixed bug where messages from appUser were not correctly marked as such after a page refresh

juliangarritano:


src/js/views/messageView.js
```diff
@@ -45,6 +45,9 @@ module.exports = Marionette.ItemView.extend({
 
     _isAppMaker: function() {
         var appMakers = this.getOption('conversation').get('appMakers');
-        return !!appMakers.get(this.model.get('authorId'));
+
+        return !!appMakers.findWhere({
```
lemieux: this is very weird...

lemieux:
:+1: we'll need to figure out why the `get(id)` didn't work at some point..

PR #29, opened 2015-07-02
F/less refactoring+styling overhaul

juliangarritano:


src/stylesheets/footer.less
```diff
@@ -0,0 +1,49 @@
...
+    background-color: white;
+    z-index: 1;
+
+    .message-input {
+        background-color: #fbfbfb;
+        color: #bdbdbd;
+        border: 1px solid #e8e8e8;
+        margin: 5px 9px;
+        padding: 0 9px;
+        border-radius: 4px 4px 4px 4px;
```
lemieux: just 4px

src/stylesheets/main.less
```diff
@@ -1,370 +1,43 @@
 @import "reset.less";
```
lemieux: I think we should put all imports under the sk-container scope (maybe not mixins) to make sure we don't interfere with the user's website styling.

lemieux:
:+1:

PR #33, opened 2015-07-06
F/logo retina

juliangarritano:


juliangarritano:
inb4 "why aren't you using css background-image?"
The main purpose of that logo (aside from increasing brand awareness) is to affect Google SEO. We need alt text, which means we need img tag.

juliangarritano:
inb4 "why not just use the 2x image?"
See a7163bb

lemieux:
:+1:

PR #34, opened 2015-07-08
Switching from margin to padding so that outerHeight calculation is correct

juliangarritano:


lemieux:
Hold on that, the whole problem is the box sizing. It's not set to content-box. It's all fixed in my email branch.

juliangarritano:
Cool! I'll close this then.

PR #35, opened 2015-07-10
Fixes for whisper support

lemieux:
The conversation was wrongly marked at initiated when it was not.

### DIFF
--- a/src/js/controllers/chatController.js
+++ b/src/js/controllers/chatController.js
@@ -153,7 +153,7 @@ module.exports = ViewController.extend({
                 .then(this._getConversation)
                 .then(this._initFaye)
                 .then(_.bind(function(conversation) {
-                    this.conversationInitiated = true;
+                    this.conversationInitiated = !conversation.isNew();
                     return conversation;
                 }, this));
         }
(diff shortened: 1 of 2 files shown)


src/js/controllers/chatController.js
```diff
@@ -153,7 +153,7 @@ module.exports = ViewController.extend({
                 .then(this._getConversation)
                 .then(this._initFaye)
                 .then(_.bind(function(conversation) {
-                    this.conversationInitiated = true;
+                    this.conversationInitiated = !conversation.isNew();
```
juliangarritano: So what does this fix, exactly?
We will get whispers from faye, now?

lemieux: Yeah, in this flow, the conversation that you get from `_getConversation` can be a new one (hence `!conversation.isNew()`) if no conversation was started. `_initFaye` is a noop in that case.

Later when you track something, it rechecks the server to see if a conversation was created and if it is the case, it connects to faye. It also fetches the conversation and the messages, so the first whispers might come throught the `fetch` call but everything else will go through faye.

lemieux: That's why it was only working on reload... the conversation was always marked as initiated so it was not refetching from the server.

juliangarritano:
:+1:

mspensieri:
:+1:

PR #36, opened 2015-07-10
fixing path for dist

daudich:
The new `demo.html` is being created in `/example` but the `/dist` file is one directory up.

### DIFF
--- a/example/template/demo.html
+++ b/example/template/demo.html
@@ -6,7 +6,7 @@
     <meta charset="utf-8">
     <title>Conversation Demo</title>
     <script src="http://code.jquery.com/jquery-1.11.3.min.js"></script>
-    <script src="/dist/WIDGET_CODE"></script>
+    <script src="../dist/WIDGET_CODE"></script>
     <script>
         /* global SupportKit: false */
 


example/template/demo.html
```diff
@@ -6,7 +6,7 @@
     <meta charset="utf-8">
     <title>Conversation Demo</title>
     <script src="http://code.jquery.com/jquery-1.11.3.min.js"></script>
-    <script src="/dist/WIDGET_CODE"></script>
```
juliangarritano: This works fine with http://localhost:8282/example/demo, but your fix will work with the file:// protocol as well.

juliangarritano:
:+1:

lemieux:
How did we never catch that?

Marc-Antoine Lemieux

2015-07-10 16:10 GMT-04:00 juliangarritano <notifications@github.com>:

daudich:
So, should I merge it then?

lemieux:
Yep go for it

juliangarritano:
We actually did catch that, I just started using localhost and forgot about it 😜

PR #38, opened 2015-07-13
Added check for dirty user prop before sending message

juliangarritano:


src/js/main.js
```diff
@@ -56,7 +56,7 @@ var SupportKit = Marionette.Object.extend({
     },
 
     _updateUser: function() {
-        return this.user.save();
+        return this.user.saveIfDirty();
```
lemieux: just override `save` if you are going to replace it

src/js/models/appUser.js
```diff
@@ -1,14 +1,16 @@
 'use strict';
 
 var _ = require('underscore'),
+    $ = require('jquery'),
     BaseModel = require('./baseModel');
 
-module.exports = BaseModel.extend({
+var AppUserModel = module.exports = BaseModel.extend({
```
lemieux: just `AppUser`

src/js/models/appUser.js
```diff
@@ -18,5 +20,28 @@ module.exports = BaseModel.extend({
             properties: {},
             conversationStarted: false
         };
+    },
+
+    isDirty: function() {
+        var hasChanged = false;
+
+        _.each(AppUserModel.EDITABLE_PROPERTIES, function(property) {
+            hasChanged = hasChanged || this.hasChanged(property);
```
lemieux: `hasChanged` is from the last set, not the last save.

src/js/models/appUser.js
```diff
@@ -18,5 +20,28 @@ module.exports = BaseModel.extend({
...
+        }.bind(this));
+
+        return hasChanged;
+    },
+
+    saveIfDirty: function() {
+        if (this.isDirty()) {
+            return this.save();
+        } else {
+            return $.Deferred().resolve().promise();
```
lemieux: that promise should replicate what  `this.save()` is return (I think it passes the model as the first parameter)

src/js/main.js
```diff
@@ -139,7 +139,7 @@ var SupportKit = Marionette.Object.extend({
                     uiText: uiText
                 });
 
-                return this.updateUser(_.pick(options, 'givenName', 'surname', 'email', 'properties'));
+                return this.updateUser(_.pick(options, AppUser.EDITABLE_PROPERTIES));
```
lemieux: great idea

src/js/controllers/chatController.js
```diff
@@ -61,6 +61,10 @@ module.exports = ViewController.extend({
         }
     },
 
+    updateUserIfDirty: function() {
```
lemieux: looks like it should be the default behavior of an `updateUser` function with an option to tell `force: true`

lemieux: The naming looks... dirty to me

dannytranlx:
:+1:

src/js/models/appUser.js
```diff
@@ -1,14 +1,25 @@
...
+    $ = require('jquery'),
+    Backbone = require('backbone'),
     BaseModel = require('./baseModel');
 
-module.exports = BaseModel.extend({
+var AppUser = module.exports = BaseModel.extend({
+    initialize: function() {
+        this._lastPropertyValues = {};
+
+        this.on('sync', function() {
```
lemieux: why not just after a successful save?

test/specs/chatController.spec.js
```diff
@@ -49,6 +49,10 @@ describe('ChatController', function() {
         renderWidgetSpy = sandbox.spy(chatController, '_renderWidget');
         receiveSpy = sandbox.spy(chatController, '_receiveMessage');
 
+        sandbox.stub(user, 'isDirty', function() {
```
lemieux: I guess it needs a test where the flag is true.

src/js/models/appUser.js
juliangarritano: Wasn't in the catalog of events when I checked backbone.
on('save')?

lemieux: I mean, you already override the save behavior, why not do that there?

PR #39, opened 2015-07-14
Forgot to bind

juliangarritano:


src/js/models/appUser.js
```diff
@@ -42,7 +42,7 @@ var AppUser = module.exports = BaseModel.extend({
             return Backbone.Model.prototype.save.apply(this, _.toArray(arguments)).then(function() {
                 this._lastPropertyValues = this.pick(AppUser.EDITABLE_PROPERTIES);
                 return arguments;
-            });
+            }.bind(this));
```
lemieux: use `_.bind` instead for compatibility with old crumbling browsers.

juliangarritano: We're not supporting old crumbling browsers though...
I hate the way _.bind looks.
If we have to go through the codebase to fix them all, I volunteer.

lemieux: I hate it too, but that's a small thing to add support for those browsers. And `_(function(){}).bind(this)` looks almost the same ;)

lemieux: Hmm nevermind... fn.bind is supported by IE9 and up.

juliangarritano: Yeah that's what I meant when I said we aren't supporting crumbling browsers. I thought you meant IE 8 :tongue:

lemieux:
Delete your branch though.

src/js/models/appUser.js
dannytranlx: The issue was with PhantomJS 1.x which runs our client-side tests. Might need to update to PhantomJS 2.x and all the related Grunt packges.

PR #41, opened 2015-07-14
Added ready promise on init

lemieux:
@alavers @jpjoyal @juliangarritano

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,10 @@ SupportKit.init({appToken: 'your_app_token'});
 ### Individual functions
 
 #### init(options)
-Initializes the SupportKit widget in the web page using the specified options
+Initializes the SupportKit widget in the web page using the specified options. It returns a promise that will resolve when the widget is ready if you need to do different things at different times.
 
 ```javascript
-SupportKit.init({
+var skPromise = SupportKit.init({
     appToken: 'your_app_token',
     givenName: 'Cool',
     surname: 'Person',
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -59,10 +59,10 @@ SupportKit.init({appToken: 'your_app_token'});
 ### Individual functions
 
 #### init(options)
-Initializes the SupportKit widget in the web page using the specified options
+Initializes the SupportKit widget in the web page using the specified options. It returns a promise that will resolve when the widget is ready if you need to do different things at different times.
```
juliangarritano: "If you need to do different things at different times"
Wut?

lemieux: Better phrasing? It allows you to keep a reference where you can always be sure the widget is init-ed

juliangarritano: "Initializes the SupportKit widget in the web page using the specified options. It returns a promise that will resolve when the widget is ready. Keeping a reference to that promise allows you to run code when you know init has completed."

I still don't like it.

williamsjj:
If code adds an init promise later after init has long since completed, will the promise fire immediately or just queue waiting for some logout/init resequence that may never happen?

juliangarritano:
As long as you keep a reference to the initial init call, any subsequent promises will fire immediately given it has been resolved.

williamsjj:
@juliangarritano Awesome. Thanks for clarifying.

lemieux:
The promise resolves at the same time `ready` is triggered. The main difference from using `ready` is that ready is a one time thing  while the promise can be `promise.then(...)` again and again.

alavers:
:+1:

PR #47, opened 2015-07-21
Fixed first message disappearing bug.

juliangarritano:
Added option to force user update. Forcing user update on appboot to ensure conversationStarted is correctly set on first message send.

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -55,14 +55,6 @@ var SupportKit = Marionette.Object.extend({
         }
     },
 
-    _updateUser: function(userInfo) {
-        userInfo && this.user.set(userInfo, {
-            parse: true
-        });
-
-        return this.user.save();
-    },
-
     init: function(options) {
         if (this.ready) {
             return;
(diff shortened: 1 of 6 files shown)


juliangarritano:
Note that this bug was only reproducible if you initialized SupportKit with ONLY the appToken.

The reason for this is that updateUser was not saving the user on init() because the model wasn't considered to be 'dirty' (none of the main properties had changed)

I fixed this by allowing a forced updateUser which is now called on init();

src/js/main.js
```diff
@@ -216,12 +220,8 @@ var SupportKit = Marionette.Object.extend({
             return $.Deferred().resolve(this.user);
         }
 
-        this.user.set(userInfo, {
-            parse: true
-        });
-
         this._throttledUpdate = this._throttledUpdate || _.throttle(this._updateUser.bind(this), 60000);
-        return this._throttledUpdate();
+        return this._throttledUpdate(userInfo);
```
lemieux: hmm I think we'll lose data if we throttle ... what happens if we do 3 update user in a row? only the last set of attributes if taken in account?

juliangarritano: What do you suggest?

juliangarritano: I guess the set being in updateUser instead sort of fixed that problem...

lemieux: yeah, so what about leaving the set in updateUser, but do a set in the other place too and call _updateUser anyway?

juliangarritano: That would work :+1:

lemieux:
@juliangarritano can you review my changes? I removed _updateUser as we discussed, but it lead to fixing the tests around that.

juliangarritano:
Remove the call to updateUser in the chatController.

Rest is good :+1:

jpjoyal:
Was this shipped ?

lemieux:
No but it's in integration. Want to ship it now?

PR #48, opened 2015-07-21
Email capture support

lemieux:
The expected behavior is the following : 

1. if the email was not provided on init and not retrieved on appboot, a notification will slide down after the first message with a link to the notification settings. 
2. if the email was provided, no notification is displayed

In the notification settings view, the behavior is the following: 

1. if the email was provided on init, the settings can't be changed. The email address will appear in a disabled input with no save button
2. if the email was not provided or was retrieve through appboot, the email address will appear in the input and the save button will appear
3. once an email address is provided, it becomes required and cannot be removed.

The email will be store in the user profile. Changes to the backend are needed to show that information using the !profile command in Slack and in other places.

I highly suggest you test that locally.


src/stylesheets/main.less
```diff
@@ -1,3 +1,4 @@
+@import "../../node_modules/bootstrap/less/mixins.less";
```
juliangarritano: Should we import this in our own mixins.less?

src/stylesheets/main.less
```diff
@@ -29,15 +29,22 @@
...
         position: relative;
         border-radius: 10px 10px 0 0;
     }
-
-    @import "reset.less";
+    @import "../../node_modules/bootstrap/less/variables.less";
+    @import "../../node_modules/font-awesome/less/font-awesome.less";
     @import "variables.less";
+    @import "reset.less";
+    @import "../../node_modules/bootstrap/less/buttons.less";
```
juliangarritano: Same here?

src/stylesheets/main.less
```diff
@@ -29,15 +29,22 @@
...
-        width: 330px;
-        height: 420px;
+        width: @widget-width;
+        height: @widget-height;
         position: relative;
         border-radius: 10px 10px 0 0;
     }
-
-    @import "reset.less";
+    @import "../../node_modules/bootstrap/less/variables.less";
```
juliangarritano: And here?

src/stylesheets/main.less
lemieux: We could but I prefer having the vendor import separate from ours.

juliangarritano:
Looks good to me.
I'll give it a try tomorrow.

src/js/views/settingsView.js
```diff
@@ -0,0 +1,124 @@
...
+        if (this.getOption('readOnlyEmail')) {
+            $el.attr('disabled', true);
+        }
+    },
+
+    onEmailSet: function(value) {
+        // http://stackoverflow.com/questions/46155/validate-email-address-in-javascript
+        var regex = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
+
+        var isValid = regex.test(value);
```
dannytranlx: Is it too soon/heavy to introduce `Backbone.validation` in the widget?

lemieux: I thought about it, but we need to keep the widget footprint as low as possible. It felt acceptable to go with that regex for now. If things are becoming more complex, I'm all for bringing in Backbone.Validation and the behavior we have in the app.

src/templates/settings.tpl
```diff
@@ -0,0 +1,22 @@
...
+        </div>
+
+        <% if (!readOnlyEmail) { %>
+            <div class="input-group">
+                <button type="button" class="btn btn-sk-primary" data-ui-save>Save</button>
+                <span class="form-message" data-ui-form-message></span>
+            </div>
+        <% } %>
+    </form>
+</div>
```
daudich: You need a newline here.

src/templates/settingsHeader.tpl
```diff
@@ -0,0 +1,3 @@
+<div class="sk-back-handle">&#10140;</div>
+Email
+<div class="sk-close-handle" data-ui-close>&times;</div>
```
daudich: And here.

src/templates/header.tpl
```diff
@@ -1,3 +1,5 @@
 <%= headerText %>
 <div id="sk-badge" data-ui-badge></div>
-<div id="sk-handle">&#10140</div>
\ No newline at end of file
+<div id="sk-notification-badge" data-ui-notification-badge><i class="fa fa-gear"></i></div>
+<div class="sk-show-handle sk-appear-hidden">&#10140;</div>
+<div class="sk-close-handle sk-close-hidden">&times;</div>
```
daudich: And here.

src/stylesheets/notifications.less
```diff
@@ -0,0 +1,22 @@
...
+            position: absolute;
+            right: 18px;
+            top: 50%;
+            font-size: 20px;
+            color: inherit;
+            text-decoration: none;
+            transform: translateY(-50%);
+        }
+    }
+}
```
daudich: Here.

src/stylesheets/settings.less
```diff
@@ -0,0 +1,45 @@
...
+                .button-size(@padding-base-vertical; @padding-base-horizontal; @font-size-base; @line-height-base; @btn-border-radius-base);
+                vertical-align: middle;
+
+                i.success {
+                    color: @brand-success;
+                }
+            }
+        }
+    }
+}
```
daudich: here.

src/templates/emailNotification.tpl
```diff
@@ -0,0 +1,2 @@
+In case we're slow to respond you can <a href="#" data-ui-settings-link>leave us your email</a>.
+<a href="#" data-ui-close class="sk-notification-close">&times;</a>
```
daudich: here.

src/templates/chat.tpl
```diff
@@ -1,5 +1,8 @@
 <div id="sk-wrapper">
     <div data-region-header class="sk-noselect"></div>
+    <div data-region-notifications class="sk-notifications">
+    </div>
+    <div data-region-settings></div>
     <div data-region-main></div>
     <div data-region-footer></div>
 </div>
```
daudich: here.

src/stylesheets/visibility.less
```diff
@@ -0,0 +1,19 @@
...
+
+&.sk-close {
+    .sk-close-hidden {
+        display: none;
+    }
+
+    .sk-close-visible {
+        display: block;
+    }
+}
```
daudich: here.

src/stylesheets/variables.less
```diff
@@ -1,7 +1,19 @@
...
\ No newline at end of file
+@conversation-height: calc(@widget-height - (@header-height + 2 * @header-padding) - @footer-height - @conversation-padding);
+
+@fa-font-path: '//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/fonts';
+
+@input-horizontal-margin: 9px;
+@input-horizontal-padding: 9px;
+
+@sk-blue: #00aeff;
+@font-size-base: 13px;
```
daudich: here.

src/stylesheets/main.less
```diff
@@ -29,15 +29,22 @@
...
+    @import "../../node_modules/bootstrap/less/buttons.less";
+    @import "forms.less";
+    @import "buttons.less";
+    @import "visibility.less";
     @import "header.less";
+    @import "notifications.less";
     @import "conversation.less";
     @import "footer.less";
+    @import "settings.less";
 }
```
daudich: Ditto.

dannytranlx:
Wow @daudich <3 I didn't want to be that nitpicky, but yeah missing some end of file newlines :grin:

lemieux:
thanks to @dannytranlx, we now have the newline police.

src/js/views/emailNotificationView.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+var Marionette = require('backbone.marionette'),
+    _ = require('underscore');
+
+var EmailNotificationTemplate = require('../../templates/emailNotification.tpl');
+
+module.exports = Marionette.ItemView.extend({
+    tagName: 'p',
```
daudich: What's this for?

src/js/views/settingsView.js
```diff
@@ -0,0 +1,124 @@
+var Marionette = require('backbone.marionette'),
+    _ = require('underscore'),
+    $ = require('jquery'),
+    bindAll = require('lodash.bindall');
```
daudich: I might just be being nit picky here but the style that I have seen so far is to have a require on individual lines with `var`s. If we want to keep this then perhaps template should also be added here?

    bindAll = require('lodash.bindall'),
    template = require('../../templates/settings.tpl');

daudich:
:+1:

src/js/views/emailNotificationView.js
lemieux: Backbone views (Marionette.ItemView inherits from it) generates a `div` by defautlt as the container for the view. `tagName` allows you to change that to something else.

src/js/views/settingsView.js
lemieux: I keep them separated for a reason, some sort of convention.

Usually I specify my requires this way:

1. One block for third-parties (backbone, underscore, jquery, etc.)
1. One block for local utils/global stuff (analytics, logging, notifications, etc.)
1. One block for views
1. One block for controllers
1. One block for models
1. One block for templates

IMO it makes it easier to read the dependencies a module has.

dannytranlx:
:ship:

juliangarritano:
Why do we need newlines?

dannytranlx:
It is usually recommend for command-line text editor to have a newline at the end of files to avoid odd behaviors like when you use `cat` on two files, they would show up on the same line, when they shouldn't be. Or when you want to jump at the end of your file, it should not have content on the last line.

http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

`git` offer a reminder when you forget your newline at the end of a file and our `.jsfmt` adds it to Javascript files usually, but not .less unfortunately.

`git diff`
image

on GitHub too
image

mspensieri:
I'm not crazy about having the email input appear only after the first message; that means the app maker has to explicitly type `/sk !profile` to be able to see it, or wait for another message in the case of email integration (which may never arrive). Wouldn't it be better to prompt before the first message, so that the initial profile dump includes the email address?

jpjoyal:
We could hold on the first message on the sdk side until we have the email.

kvgauthier:
Asking for the email upfront is we think a subpar experience for the end-user. I don't think it would be a good idea to hold the first message on the SDK since it would a) delay the message and b) could lose the message altogether if the tab is closed.

All in all, once true email notification is implemented for end-users, it will become mostly irrelevan for app makers to have the email or not in the first message.

alavers:
I have some ideas here, lets move to a better medium like slack

williamsjj:
I agree with @kvgauthier that delaying the message is a subpar experience for the user. I'd rather then receive a quick response if they're anonymous than know always who they are. 

Why not make the email box always visible if not yet supplied and notated as optional?

lemieux:
@alavers @juliangarritano this should be good for a last pass on it. 

Throttling is back and it's directly on the appUser save method.

src/js/views/settingsView.js
alavers: I've seen @lemieux and @juliangarritano use this multi line var syntax. I don't like it, this guy explains why: (http://benalman.com/news/2012/05/multiple-var-statements-javascript/)

I've refrained from commenting (until now) because its not a huuuge deal, If it was just my opinion I was arguing I'd let it slide. However, the separate var style, eg:

```
var Marionette = require('backbone.marionette');
var _ = require('underscore');
```

is far more common across the SK codebase, and because style *consistency* across code bases is far more important than any one person's opinion about style I would suggest that supportkit-js should follow suit. You can use newlines to separate 3rd party from views from controllers etc.

I would further suggest that if possible we automate confirming code with tools like jsfmt @dannytranlx is there an option for this?

lemieux: Sold, I'll adjust .jsfmt to automate that.

lemieux:
It should be all good now. I reduced the number of things we were adding to the stylesheet to strictly what we are using and I also applied modifications on code style based on @alavers comments.

.jsfmtrc
```diff
@@ -302,7 +302,9 @@
 
   "plugins": [
     "esformatter-quotes",
-    "esformatter-braces"
+    "esformatter-braces",
+    "esformatter-var-each",
+    "esformatter-dot-notation"
```
alavers: Nice, I'll add these in the other repo

alavers:
@williamsjj for context, note that this PR is an incremental change that's part of a feature that is still under active development, in other words, it's not the final state we want for SupportKit on the web.

The current flow here is that we only prompt for the user's email *after* the first message. We don't want a form to be the first thing a user sees, even if its just one field. We want to keep the chat UI as inviting as possible. Important to note that this current state only makes sense for the slack integration where you have the option to do `/sk !profile` to see the email the user left for you. There's no equivalent way (yet) to do this profile lookup with the Zendesk integration but we are working on fixing this.

This is why we have it disabled by default, and why you have to explicitly opt-in with the `emailCaptureEnabled` option.

src/js/main.js
```diff
@@ -312,6 +312,10 @@ var SupportKit = Marionette.Object.extend({
 
             this.ready = false;
         }
+    },
+
+    __showNotification: function() {
```
alavers: nit, is that two underscores? if so is that on purpose?

lemieux: not on purpose, good catch

.csscomb.json
```diff
@@ -0,0 +1,329 @@
...
+            "box-decoration-break",
+            "-webkit-box-shadow",
+            "-moz-box-shadow",
+            "box-shadow",
+            "filter:progid:DXImageTransform.Microsoft.gradient",
+            "-ms-filter:'progid:DXImageTransform.Microsoft.gradient",
+            "text-shadow"
+        ]
+    ]
+}
```
dannytranlx: Nice!

juliangarritano:
This looks good to me!
:+1:

src/js/controllers/chatController.js
```diff
@@ -76,25 +77,77 @@ module.exports = ViewController.extend({
...
-                text: text
-            }, {
-                success: messageDeferred.resolve,
-                error: messageDeferred.reject
-            });
 
-        });
+        conversationDeferred.then(_.bind(function(conversation) {
+            // update the user before sending the message to ensure properties are correct
+            this.user.save({}, {
```
juliangarritano: Please make sure this doesn't cause regression on first user message disappearing when SupportKit is initialized with only the appToken

src/js/controllers/chatController.js
```diff
@@ -1,22 +1,25 @@
 'use strict';
```
juliangarritano: This file is getting pretty huge.
We might want to consider trying to split it up somehow.
For now it's fine, though.

lemieux: I agree, it's mostly the real "main" controller. My plan at some point is to create a controller to manage the different headers (because we might have more than 2 at some point) and to move the chat input to it's own controller that would also take care of upload and different input when we do that feature. 

I'd say next time we need more views, we should break up this controller in smaller pieces.

src/js/controllers/chatController.js
lemieux: I'll try to figure a way to write tests for those cases.

src/js/main.js
```diff
@@ -130,9 +140,18 @@ var SupportKit = Marionette.Object.extend({
 
                 endpoint.appUserId = res.appUserId;
 
+                // if the email was passed at init, it can't be changed through the web widget UI
+                var readOnlyEmail = !_.isEmpty(options.email);
+
+                var emailCaptureEnabled = options.emailCaptureEnabled && !readOnlyEmail
+
```
mspensieri: Too many newlines!

src/js/models/appUser.js
```diff
@@ -19,6 +27,40 @@ module.exports = BaseModel.extend({
             properties: {},
             conversationStarted: false
         };
+    },
+
+    isDirty: function(attributes) {
+        attributes || (attributes = {});
+
+        var comparableAttributes = _.extend({}, this.attributes, attributes);
+
```
mspensieri: Here too

src/js/controllers/chatController.js
mspensieri: I think we need a way to bypass the throttle on this call - unsaved user details should _always_ be uploaded before a message, regardless of when the previous upload happened

lemieux: I'll call `user._save` then. It's the unthrottled one.

dannytranlx:
:rocket:

PR #52, opened 2015-07-24
Release v0.2.17

lemieux:
Bug fixes for word-wrap on FF and IE.

### DIFF
--- a/release_notes/v0.2.16
+++ b/release_notes/v0.2.16
@@ -1,2 +1,0 @@
-# Bug Fixes
-- Fixed disappearing first message bug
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "backbone-computedfields": "0.0.10",
     "backbone.marionette": "^2.4.1",
     "backbone.stickit": "^0.9.2",
+    "bootstrap": "^3.3.5",
```
jpjoyal: did you check of this is bringing conflicts with sites that already include bootstrap ? or some other version of bootstrap

lemieux: We're not using the JS, just the CSS part. And we are using only some parts of it and it's all scoped under sk-container.

PR #54, opened 2015-07-28
setting modified user properties on updateUser

pranavguptarulz:


lemieux:
I'll close this one because we decided to remove that flag in #55  and reimplement a better way to do this later.

Thanks again for this!

PR #55, opened 2015-07-28
Remove isDirty check on user model

### LINKED ISSUE #53: SupportKit.updateUser doesn't work
We at [razorpay](https://razorpay.com) are relying on `SupportKit.updateUser` for adding email and contact info. But this doesn't seem to work since last release.

`SupportKit.user.set('email', 'foo@bar.com')` followed by `SupportKit.user.save()` is working though, and we're falling back to it for now.

lemieux:
This should fix #53. That check is removed for now, but it should be back when we reintroduce throttling.

@jpjoyal @alavers @juliangarritano

### DIFF
--- a/src/js/controllers/chatController.js
+++ b/src/js/controllers/chatController.js
@@ -79,8 +79,7 @@ module.exports = ViewController.extend({
         conversationDeferred.then(function(conversation) {
             // update the user before sending the message to ensure properties are correct
             this.user.save({}, {
-                wait: true,
-                parse: true
+                wait: true
             }).then(function() {
                 conversation.get('messages').create({
                     authorId: endpoint.appUserId,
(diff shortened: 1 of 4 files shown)


src/js/controllers/chatController.js
```diff
@@ -79,14 +79,17 @@ module.exports = ViewController.extend({
         conversationDeferred.then(function(conversation) {
             // update the user before sending the message to ensure properties are correct
             this.user.save({}, {
-                wait: true
+                wait: true,
+                parse: true
```
alavers: Is this option necessary?

alavers:
:+1:

src/js/controllers/chatController.js
lemieux: good point, not sure we need it on that one. We have a custom parse function, because the server sometimes returns only the id.

PR #59, opened 2015-07-29
Support for gzip

### LINKED ISSUE #58: Enable gzip on supportkit CDN
The current size of `supportkit.min.js` is 215kb, which will be going up once #48 is merged. Unfortunately, supportkit is the slowest and the largest JS file on our site, and your cloudfront config doesn't have gzip enabled.

Please enable gzip: http://checkgzipcompression.com/?url=https%3A%2F%2Fcdn.supportkit.io%2Fsupportkit.min.js

lemieux:
As per AWS documentation, we need to create a second compressed file and upload it to S3 with gzip encoding. 

This would partly fix #58. I think we still need some server-side/CDN config to serve the right file.

@jpjoyal  @alavers @juliangarritano

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -71,7 +71,7 @@ module.exports = function(grunt) {
             },
             dist: {
                 src: 'dist/supportkit.js',
-                dest: 'dist/supportkit.min.uncompressed.js'
+                dest: 'dist/supportkit.min.js'
             }
         },
 
@@ -145,8 +145,8 @@ module.exports = function(grunt) {
                 options: {
                     mode: 'gzip'
                 },
-                src: 'dist/supportkit.min.uncompressed.js',
-                dest: 'dist/supportkit.min.js'
+                src: 'dist/supportkit.min.js',
+                dest: 'dist/supportkit.min.compressed.js'
             }
         },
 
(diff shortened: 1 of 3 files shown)


lemieux:
Hmmm apparently that way of do it is not working for S3 since it doesn't do detection.. we would need to only upload the gzipped file to the CDN (http://stackoverflow.com/questions/22196236/can-someone-walk-me-through-serving-gzipped-files-from-cloudfront-via-s3-origin)

I'm not sure if it can cause problems down the road.

juliangarritano:
This looks good, but how can we test it safely?

juliangarritano:
Can we upload a test copy to the CDN and remove it once we verify the gzipped version is functioning correctly?

juliangarritano:
Also why do we need the non-gzipped version on the CDN at all?

lemieux:
yeah good idea. I'll leave the supportkit.min.js file uncompressed and have a supportkit.min.compressed.js file with it in a first release to test it. If it goes well, I'll do a second release with a compressed supportkit.min.js and an uncompressed supportkit.min.uncompressed.js just to be safe in case someone needs the uncompressed one.

juliangarritano:
Sounds good to me
:+1:

PR #61, opened 2015-07-29
Support for gzip, part 3

lemieux:
Third release for #59.


juliangarritano:
👍

lemieux:
I'll merge that anyway. It will have the stuff needed for gzip support when we figure out server-side.

It doesn't change current behavior. It just adds a `supportkit.min.js.gz` file which is the pre-compressed file on S3 and Cloudfront.

PR #69, opened 2015-08-05
Race condition on Faye connection

lemieux:
We were not using the faye promise after connecting which seems to cause a race condition when we create a collection then connect faye. This fixes it.


mspensieri:
:+1:

PR #70, opened 2015-08-05
Release v0.2.23

lemieux:
# Bugs
- Fixed a race condition where faye didn't connect in time to get a whisper after the first user message.


lemieux:
I'm holding this merge because CircleCi is not picking up new commits for builds and would most likely not trigger a deployment. If they don't fix that really soon, I'll proceed to a manual deployment.

PR #74, opened 2015-08-06
Reset the ready promise on logout

lemieux:
That promise was not reset on logout which made subsequent call weird after a logout/init.


mspensieri:
:+1:

PR #75, opened 2015-08-06
Close the settings view on save

lemieux:
This behavior was asked by a couple of users and even by us during review. This streamlines the email capture experience for the end user and get them back to the conversation as fast as possible.

### DIFF
--- a/src/js/controllers/settingsController.js
+++ b/src/js/controllers/settingsController.js
@@ -29,11 +29,11 @@ module.exports = ViewController.extend({
     },
 
     onSettingsSave: function() {
-        this.listenToOnce(this.model, 'sync', function() {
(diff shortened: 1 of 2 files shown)


src/js/controllers/settingsController.js
```diff
@@ -30,13 +30,17 @@ module.exports = ViewController.extend({
 
     onSettingsSave: function() {
         this.listenToOnce(this.model, 'sync', function() {
```
mspensieri: Could you register this in the `isDirty` branch of the `if` statement instead?

lemieux: Yeah it makes more sense

mspensieri:
:+1:

PR #77, opened 2015-08-07
First message whisper fix : take 2

lemieux:
The promise chain in `sendMessage` was complicated it seemed to be wrong. I rewrote it in a much cleaner way.

### DIFF
--- a/src/js/controllers/chatController.js
+++ b/src/js/controllers/chatController.js
@@ -63,53 +63,51 @@ module.exports = ViewController.extend({
     },
 
     sendMessage: function(text) {
-        var conversationDeferred = $.Deferred();
-        var messageDeferred = $.Deferred();
+        return $.Deferred().resolve().then(function() {
+            var promise = $.Deferred();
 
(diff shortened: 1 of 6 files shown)


src/js/controllers/chatController.js
```diff
@@ -63,53 +63,51 @@ module.exports = ViewController.extend({
     },
 
     sendMessage: function(text) {
-        var conversationDeferred = $.Deferred();
-        var messageDeferred = $.Deferred();
+        return $.Deferred().resolve().then(function() {
+            var promise = $.Deferred();
 
-        if (this.conversation.isNew()) {
```
mspensieri: At which point was the actual breakage in the old chain?

mspensieri:
Looks good to me :+1:

src/js/controllers/chatController.js
lemieux: _initFaye and createConversation seemed to be execute at the same thing, which caused faye to be subscribed too late

PR #82, opened 2015-08-18
Render placeholder content so bots can crawl links

chris-rudmin:
@alavers @mspensieri @lemieux @juliangarritano @edchan77

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -17,6 +17,8 @@ var Event = require('./models/event');
 /*jshint +W079 */
 var Rule = require('./models/rule');
 var AppUser = require('./models/appUser');
+var ChatView = require('./views/chatView');
+var ConversationView = require('./views/conversationView');
 var ChatController = require('./controllers/chatController');
 var Conversations = require('./collections/conversations');
 var endpoint = require('./endpoint');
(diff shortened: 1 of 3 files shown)


alavers:
:+1:

src/js/main.js
```diff
@@ -53,6 +55,7 @@ var SupportKit = Marionette.Object.extend({
         bindAll(this);
         this._readyPromise = $.Deferred();
         this._conversations = new Conversations();
+        this._placeHolder = $();
```
lemieux: I don't think this is needed (see comments below)

src/js/main.js
```diff
@@ -274,17 +271,28 @@ var SupportKit = Marionette.Object.extend({
...
         }
     },
 
+    // Add placeholder to support back links to supportkit
+    _renderPlaceholder: function() {
+        $(_(function() {
+            if (!this.ready) {
+                var placeHolder = new ChatView().render();
+                placeHolder.main.show(new ConversationView());
+                this._placeHolder = $(placeHolder.el);
```
lemieux: I'd do 

```
this._placeHolder = new ChatView();
this._placeHolder.main.show(new ConversationView());
this._placeHolder.render().$el.addClass('placeHolder').appendTo('body');
```

in the current form, we are leaking a ChatView instance and a ConversationView instance because we only remove the element from the DOM and not disposing of those views.

lemieux:
some comments, but other than that :+1:

chris-rudmin:
@lemieux Is using`destroy()` good solution?

lemieux:
yup :+1:

PR #85, opened 2015-08-24
Icon in header use header color

lemieux:


lemieux:
Icon color should be lighter than the text... this PR is not needed.

PR #88, opened 2015-09-02
Fix banner version

lemieux:
It was always one version behind because it took the one from package.json instead of the new one.


mspensieri:
:+1:

PR #89, opened 2015-09-04
Fix for the version in CDN banner and removed compression

lemieux:
#88 fixed the banner for the dist files in the releases, but removed the version from the file on the CDN. I added `globalVersion` in the grunt config and it falls back to the current version in package json.

When we do a version bump, `globalVersion` is increased and then used when uglifying the built file. The deploy task was not setting `globalVersion`. This now works because we expect `globalVersion` to be the right one when we deploy.

I also removed the tasks about compression since it was not used and we'll just change CDN soon to take care of that.


mspensieri:
:+1:

PR #93, opened 2015-09-10
Remove placeholder leftovers

lemieux:
Thanks @chris-rudmin for pointing those out :)


dannytranlx:
Technically we could revert the whole #82 or double check the diff with it.

:+1:

PR #97, opened 2015-09-11
Leverage window.fetch to talk to the server instead of $.ajax

lemieux:
This is step one for the login API. Since we had conlict problems using $.ajax to set headers for authorization when appMakers are also using them, I replaced our whole ajax layer with window.fetch polyfill from Github.

~~I also included some goodies like a Babelify transform in browserify to let us use es6 stuff like Promises and polyfill everything.~~ I added a polyfill for es6-promises. I'm slowly trying to replace stuff relying on jQuery by native equivalents.

I also had to rewrite the test server mock since it was relying on xhrs to do that. Apparently, it doesn't play nice with window.fetch.

~~I created the PR now to let you take a look and comment while I test this against other browsers than Chrome and PhantomJS.~~ It was tested against Chrome, IE10 and Firefox.


src/js/controllers/chatController.js
```diff
@@ -62,51 +62,46 @@ module.exports = ViewController.extend({
...
-                    error: promise.reject
-                });
-            } else {
-                promise.resolve(this.conversation);
-            }
-
-            return promise;
-        }.bind(this))
+        var self = this;
+        return new Promise(function(resolve, reject) {
```
juliangarritano: Euh is this ES6?

lemieux: yep, I added the ES6-promise polyfill

mspensieri:
:+1:

src/js/main.js
```diff
@@ -67,12 +67,12 @@ var SupportKit = Marionette.Object.extend({
         }
 
 
-        if(/lebo|awle|pide|obo|rawli/i.test(navigator.userAgent)) {
+        if (/lebo|awle|pide|obo|rawli/i.test(navigator.userAgent)) {
```
juliangarritano: Do we not want to bring back what Chris had? Do we know if this method works correctly with Google's bot?

lemieux: Unrelated to the PR, but still a good question. My guess is that it's probably as good as what Chris had. Instead of injecting the whole widget it only injects the link. The regex should cover any user agent that has those string (Googlebot, crawler, spider, robot, crawling) which covers mostly all civilized crawlers.

juliangarritano:
As long as everything works
:+1:

src/js/controllers/chatController.js
```diff
@@ -62,51 +62,46 @@ module.exports = ViewController.extend({
...
-                this.conversation = this.collection.create(this.conversation, {
-                    success: promise.resolve,
-                    error: promise.reject
-                });
-            } else {
-                promise.resolve(this.conversation);
-            }
-
-            return promise;
-        }.bind(this))
```
Mario54: Curious, why remove the `.bind(this)` and use a `self` variable instead?

lemieux: Good question. I have strong feelings against using `var self = this`, but in this case it would mean adding `.bind(this)` on all Promise functions and it looked sooooo ugly and unreadable that I settled for a small `var self = this` at the top of the chain.

I can't wait to have ES6 support where fat arrows function will take care of that!

I started this PR by adding babel in the build process but it added too much weight to our lib.

Mario54: Ah ok. Makes sense.

But yeah, if we could use ES6, this would be a good place to use arrow functions.

Mario54: It added weight to the distributed files? How much?

lemieux: a mere 60kb :/ ..

lemieux: Babel polyfills *everything* to make it on par with a fully ES6 compliant environment. It's a bit overkill if you don't need everything.

Mario54: Ah, I see.

PR #98, opened 2015-09-16
Removed browserify-shim

lemieux:
Browserify-shim is aliasing all the `require('jquery')` calls to use `window.$`, but jQuery is still bundled in the final package. Until we figure out how to remove it from there or we just get rid of it, this will at least make use of it.

Since the deployment process is broken in CircleCI, I'll do a manual deployment.


jpjoyal:
:+1:

mspensieri:
:+1:

PR #101, opened 2015-09-22
Web login

lemieux:
It should cover the API spec for the login.


src/js/controllers/chatController.js
```diff
@@ -352,14 +353,27 @@ module.exports = ViewController.extend({
 
     _getLatestReadTime: function() {
         if (!this.latestReadTs) {
-            this.latestReadTs = parseInt(cookie.parse(document.cookie).sk_latestts || 0);
+            var key = SK_LATEST_TS + '_' + endpoint.appUserId;
```
mspensieri: Read state should be tracked by `userId`. `appUserId` can pretty much change at any time

src/js/main.js
```diff
@@ -63,131 +75,128 @@ var SupportKit = Marionette.Object.extend({
...
-        });
+            return this.login(options.userId, options.jwt);
+    },
 
-        if (typeof options === 'object') {
-            endpoint.appToken = options.appToken;
-            endpoint.jwt = options.jwt;
-            if (options.serviceUrl) {
-                endpoint.rootUrl = options.serviceUrl;
+    login: function(userId, jwt) {
```
mspensieri: When a user transitions from no userId to having a userId, we should clear the last read timestamp for the anonymous user, since they will be upgraded into the new `userId` and the conversation will be cleared

test/mocks/supportKitServer.mock.js
```diff
@@ -21,9 +21,22 @@ module.exports = BaseServerMock.extend({
...
+            'POST', /\/api\/appboot/, function(xhr) {
+                var body = JSON.parse(xhr.requestBody);
+                var data = _.extend({}, appData);
+                if (body.userId) {
+                    _.extend(data, {
+                        appUserId: body.userId
+                    });
+
+                    _.extend(data.appUser, {
+                        _id: body.userId
```
mspensieri: What's the intent here? `userId` is not analogous to `appUser._id`

src/js/controllers/chatController.js
lemieux: Ok I guess I got mixed up between userId and appUserId

src/js/main.js
lemieux: How would I know it's transitionning from an anonymous or a proper using? both have ids... I guess I could put a flag on the user when they log in without a userID

test/mocks/supportKitServer.mock.js
lemieux: Mix up

src/js/main.js
lemieux: I guess one will have a userId and the other will not... again.. mixup with appUserId :(

src/js/controllers/chatController.js
lemieux: fixed.

src/js/main.js
lemieux: fixed.

test/mocks/supportKitServer.mock.js
lemieux: fixed.

src/js/controllers/chatController.js
```diff
@@ -21,6 +23,8 @@ var EmailNotificationView = require('../views/emailNotificationView');
 var ChatInputController = require('../controllers/chatInputController');
 var SettingsController = require('../controllers/settingsController');
 
+var SK_LATEST_TS = 'sk_latestts';
```
alavers: '`sk_latest_ts'`? </ocd>

src/js/controllers/chatController.js
```diff
@@ -1,3 +1,4 @@
+/* global Promise:false */
```
alavers: Why are these necessary?

src/js/faye.js
```diff
@@ -3,33 +3,26 @@
...
-    // that only has `then(resolve, reject)` and `all` methods.
-    var deferred = $.Deferred();
-
-    faye.subscribe('/conversations/' + conversationId, function(message) {
+    return faye.subscribe('/conversations/' + conversationId, function(message) {
         vent.trigger('receive:message', message);
-    }).then(function() {
-        deferred.resolve(faye);
-    }, function(err) {
-        console.error('Faye subscription error:', err && err.message);
```
alavers: Why remove this error message wrapper?

src/js/main.js
```diff
@@ -51,8 +59,12 @@ var SupportKit = Marionette.Object.extend({
 
     initialize: function() {
```
alavers: The same API has both init and initialize ಠ_ಠ

I'd much prefer to have our SupportKit global stay "clean" and expose only public API, without our internal _foo stuff and Marionette.Object stuff.

Eg, have this guy be `var Main = Marionette.Object.extend({`

and add a separate supportkit.js wrapping `Main` that exposes `init` `login` etc..

src/js/controllers/chatController.js
lemieux: I don't make the rules, jshint is :p ... I'll move that to the jshint config instead.

src/js/controllers/chatController.js
lemieux: good point

src/js/main.js
```diff
@@ -77,117 +88,122 @@ var SupportKit = Marionette.Object.extend({
...
+                platform: 'web'
+            }
+        };
 
-                var RuleCollection = Backbone.Collection.extend({
-                    model: Rule
-                });
+        if (userId) {
+            data.userId = userId;
+            endpoint.userId = userId;
```
alavers: Shouldn't endpoint.userId be unset if userId is unspecified?

alavers: Also, should it be reverted to some previous value if appboot call returns 4xx?

lemieux: it's done in `endpoint.cleanState()`

src/js/main.js
lemieux: fixed.

src/js/faye.js
lemieux: es6-promise seems to log errors to the console and reject the promise. It felt useless to have this here.

src/js/main.js
lemieux: As for reverting, I felt like the login action would always be logout first, then login. If login fails, log out still happened.

src/js/controllers/chatController.js
lemieux: fixed

alavers:
:+1:

PR #103, opened 2015-09-25
Fix appBoot and firstMessage whisper bug

lemieux:
There was a bug in the way the response from the server was handling cause a `track` to never be successful (the server got the info and all but the client always had an error).

### DIFF
--- a/src/js/controllers/chatController.js
+++ b/src/js/controllers/chatController.js
@@ -36,7 +36,6 @@ module.exports = ViewController.extend({
         this.isOpened = false;
         this.user = this.getOption('user');
         this.uiText = this.getOption('uiText') || {};
-        this.conversationInitiated = false;
         this.collection = new Conversations();
     },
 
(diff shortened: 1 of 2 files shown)


src/js/utils/api.js
```diff
@@ -26,6 +26,10 @@ function status(response) {
 }
 
 function json(response) {
+    if (response.status === 202 || response.status === 204) {
```
mspensieri: what does `response.json()` return if not `undefined` or `null`? Or does it throw?

lemieux: Pretty sure it was throwing, but now I'm just trying to test it again and I can't figure a call we do an end up with a 204

lemieux:
After testing with @Mario54 , this PR doesn't fix the first user message bug at all. I'll have to work on that.

src/js/utils/api.js
lemieux: Yeah it was throwing. I tested it with tracking an event for a whisper.

src/js/controllers/chatController.js
```diff
@@ -197,17 +203,14 @@ module.exports = ViewController.extend({
     _initConversation: function() {
         var promise;
 
-        if (this.conversationInitiated) {
+        if (this.collection.length > 0 && !this.collection.at(0).isNew()) {
```
lemieux: This is the key change that fixed the first message disappearance. It relied on `this.conversationInitiated` that was set to true further down the promise chain. I'm pretty there was a race condition when we tracked an event and/or update the user and the conversation is started there too.

lemieux:
@Mario54 tested it and the bug seems to be gone. I was able to reproduce it very easily and it seems to be gone too.

spasiu:
:+1:

mspensieri:
:+1:

PR #105, opened 2015-09-29
Actions and links support 

lemieux:
This PR was branched from `b/conversation-first-message`.

This brings support for actions and link detection in the web sdk. Currently the link detection is fairly simple, it detects anything that might look like an URL (protocol://uri). We should make it more versatile in the future to detect things like email addresses and generate a proper mailto href.


src/js/views/messageView.js
```diff
@@ -26,11 +29,23 @@ module.exports = Marionette.ItemView.extend({
                 }
             },
             '@ui.message': {
-                observe: 'text',
-                update: function($el, text) {
-                    var escapedText = $('<div/>').text(text).html().replace(/\n/g, '<br />');
+                observe: ['text', 'actions'],
+                update: function($el, values) {
+                    if (values[0].trim().length > 0) {
```
mspensieri: Shouldn't there be an `else` branch here?

lemieux: No, if no text was provided (just an action), we don't care about rendering anything. The `span` will remain empty.

mspensieri: I guess I'm still in the iOS mindset of reusing views. Carry on

mspensieri:
:+1:

src/js/views/messageView.js
lemieux: Not sure what you mean by reusing views, but this binding would allow the message to be rendered without rerendering the whole view if the text changed somehow (which should never happen in this case)

dannytranlx:
:100:

PR #106, opened 2015-09-30
Fixed some bugs

juliangarritano:
Now blurring button after mouseup.
Now wrapping buttons that span more than one line.

### DIFF
--- a/src/stylesheets/conversation.less
+++ b/src/stylesheets/conversation.less
@@ -1,15 +1,14 @@
 [data-region-main] {
-    @media (max-width: @screen-sm-min) {
+    @media (max-width:@screen-sm-min) {
         height: @main-region-height-mobile;
     }
-
     #sk-conversation {
         padding: @conversation-padding 18px 0 13px;
         height: @conversation-height;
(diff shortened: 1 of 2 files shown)


lemieux:
:+1:

mspensieri:
:+1:

dannytranlx:
:+1:

src/stylesheets/conversation.less
```diff
@@ -190,5 +182,4 @@
             }
         }
     }
-}
-
+}
```
dannytranlx: @lemieux Doesn't the new csscomb add those usually? Are we all using the same formatter?

lemieux: ask @juliangarritano if he is using it first :p

juliangarritano: I am not using csscomb/
Dafuq is that?

juliangarritano: Should I be using csscomb?

juliangarritano: I don't really like the spacing it does much

Issue #73, opened 2015-08-06
Check if team members are active in the support kit channel


ahansson89:
I found that it is possible to check if users are active.

Could we somehow have a method checking if there is any users active in our Slack channel, then load the widget if there are any team members active or give the user a notice that we will answer through email instead of the widget?

XavierOlland:
I was planning on developing such a feature : 
- Showing the widget only if someone is there
- Have a another default behaviour like a contact form if no-one's there

I should work on this next week, if I'm able to produce something OK I'll publish it.

ahansson89:
Awesome! Looking forward to it!

lemieux:
Hey @ahansson89 and @XavierOlland , can you elaborate on your solutions? This is an interesting feature for sure!

XavierOlland:
Sorry, some delay... We're launching our product by the end of the month...

So, here is the concept:
When the page launch, SupportKit check if users subscribed to the support channel are connected.
If one or more people are connected, it shows the live chat.
if no one is, then live chat doesn't appear.

Then, assuming I can deeper in the code the live chat panel would host a contact form (name, email message).

@ahansson89 if you have found pieces of code for the connection check, it might save me some time.

pudgereyem:
This would be really useful.
I would also like to set which hours of the day the team is available.
So, basically, if not during "work hours", show contact form instead.

XavierOlland:
Working hours might be a second option and people should be allowed to choose one or both. Showing the contact form is something I had in mind, it's the way it's done in TawkTo that I tried previously. But since building a form on a website takes approximately 10 minutes. I will first concentrate on the on/off feature. Contact form would be the evolution of this work, adding a new option.

XavierOlland:
And here it is ! I did 2 scripts, one in pure javascript, and an AngularJS + Restangular. 
Please report any issue and ask if you need any more info

https://github.com/ZeroKeyboard/supportkit-js/tree/integration/integration

More to come, I'll start working on the geoIP integration

lemieux:
Thanks for this @XavierOlland. Feel free to use this on your website, but we don't feel comfortable pulling this into the main repo, as there are big security concerns associated with publishing your Slack API token in your website's source code (anyone who looks in the code can see your token, and then do anything they like to your Slack instance).

I'll close this issue since there's not much more to do.

XavierOlland:
[...]

PR #110, opened 2015-10-06
Slack presence and geo info

XavierOlland:
This commit offers 2 integration files (pure JS and AngularJS) to check if someone is online before triggering the live chat.
the simple JS files also retrieve geolocation information from users.


lemieux:
Hey @XavierOlland, 
as discussed in #73 , we won't pull this in the main repo. However, we'll make sure to direct people to your fork if our users ask about something similar.

Thanks again!

PR #116, opened 2015-10-23
Upgrade guide

lemieux:
There it is.

@alavers @jpjoyal

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,7 +2,7 @@
 
 SupportKit is now Smooch!
 
-To reflect this transition, we bumped up the major version to 2.0. We introduced a lot of breaking changes in the internals of the library to let it hit our brand new public API.
+With Smooch, the internals of the SDK have changed. It is now using our [public API](http://docs.supportkit.io/rest/) and as such we've bumped the major version to 2.0.
 
 ## CDN
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,7 +2,7 @@
 
 SupportKit is now Smooch!
 
-With Smooch, the internals of the SDK have changed. It is now using our [public API](http://docs.supportkit.io/rest/) and as such we've bumped the major version to 2.0.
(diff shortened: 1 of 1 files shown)


jugarrit:
image

jugarrit:
http://i.imgur.com/RG0BS1U.gif

UPGRADE.md
```diff
@@ -0,0 +1,37 @@
+# Upgrade guide to Smooch
+
+SupportKit is now Smooch!
+
+To reflect this transition, we bumped up the major version to 2.0. We introduced a lot of breaking changes in the internals of the library to let it hit our brand new public API.
```
alavers: Announcing breaking changes sounds harsh and might discourage people from adopting it. In reality, none of the SDK methods have changed, so moving to smooch should be easy for most integrators. Avoiding the term breaking changes might communicate this better. EG:

```With Smooch, the internals of the SDK have changed. It is now using our public API (<--make this a hotlink) and as such we've bumped the major version to 2.0```

Comment (lemieux) [inline UPGRADE.md]: done.

File: UPGRADE.md [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+
+The most notable change is that you no longer use the `SupportKit` global variable, but the new `Smooch` global.
+
+```javascript
+var Smooch = require('smooch');
+
+Smooch.init({appToken: 'your_app_token'});
+```
+
+The rest is pretty much the same! We didn't change the CSS classes, so your customization should be safe.
```
jpjoyal: I'm hesitant about not changing those. those will be the only sk remaining in the web sdk

lemieux: What's the new prefix?

And yeah, it wouldn't break anything.. it would just be part of the upgrade process.

dannytranlx: I don't see the issue of keeping `sk-`, we have a few over the other SDKs too!

spasiu: :+1:

lemieux: talked with @jpjoyal , no change needed

dannytranlx:
:+1:

PR #122, opened 2015-11-03
Add try catch to handle local storage errors

dannytranlx:
@lemieux @jugarrit @jpjoyal @spasiu 

Fixes the issue crashing Safari in private mode since the localStorage is disabled. It is showing up in `window`, throws when we try to `setItem`.

### DIFF
--- a/src/js/controllers/chatController.js
+++ b/src/js/controllers/chatController.js
@@ -7,6 +7,7 @@ var ViewController = require('view-controller');
 var endpoint = require('../endpoint');
 var vent = require('../vent');
 var faye = require('../faye');
+var storage = require('../utils/localStorage');
 
 var ChatView = require('../views/chatView');
 var HeaderView = require('../views/headerView');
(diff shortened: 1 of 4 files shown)


src/js/controllers/chatController.js
```diff
@@ -331,14 +331,21 @@ module.exports = ViewController.extend({
...
+            latestReadTs = 0;
         }
         return latestReadTs;
     },
 
     _setLatestReadTime: function(ts) {
         var key = this._getLatestReadTimeKey();
-        localStorage.setItem(key, ts);
+
+        try {
```
alavers: Might be worth wrapping localStorage.setItem in a util method so you don't have this duplicated code

alavers:
Rest looks :+1:

src/js/main.js
```diff
@@ -5,13 +5,14 @@ var Backbone = require('backbone');
 var _ = require('underscore');
 var $ = require('jquery');
 var uuid = require('uuid');
-var urljoin = require('url-join');
+var urljoin = require('urljoin');
```
dannytranlx: @lemieux This is weird, how did your build pass before?

jpjoyal:
:+1:

src/js/main.js
lemieux: very good question

src/js/utils/localStorage.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+function setItem(key, value) {
+    try {
+        localStorage.setItem(key, value);
```
lemieux: can you just fallback to an in memory object?

dannytranlx: Done!

lemieux:
:+1:

PR #124, opened 2015-11-09
Update package.json

jmendiara:
`jstify` and `node-lessify` are not used at runtime, but when included in dependencies section, are downloaded by npm, making a bigger bundle when using browserify.


palmerabollo:
+1, smooch.min.js is currently around 300K (over 110K compressed)

williamsjj:
intercom.io's chat weighs in at 150k compressed, so anything that reduces this helps (Zopim runs ~200k).

lemieux:
Good catch @jmendiara! They should definitely be in devDependencies. However it doesn't change the file size since browserify doesn't pull them in (they are not required anywhere in the code, there are just transformations applied during the build process).

FYI, Smooch is currently 116kb minified and gzipped and 324kb minified and non-gzipped. We are still quite small compare to other widget available. 

In comparison, Intercom widget is 182kb minified and gzipped, but 596kb minified and non-gzipped. (reference : https://dmmj3mmt94rvw.cloudfront.net/assets/js/hosted/intercom2-cf87dcc89c919a4f916cae61de959f84.js)

Zopim widget is 207kb minified and gzipped, but 893kb minified and non-gzipped (reference: https://v2.zopim.com/bin/v/widget_v2.118.js)

williamsjj:
Awesome. Let's get it to 50k. ;)

PR #126, opened 2015-11-09
Add support for link in brackets

dannytranlx:
Links we get from the `node-html-to-text` we use to parse integrations messages returns this syntax `check this link [https://link.com]` which isn't handle by the current regex. I've added the possibility to the regex to have a `[` at the beginning.

image
image


lemieux:
:+1:

jugarrit:
:+1:

mspensieri:
:+1:

PR #127, opened 2015-11-16
Fix supportkit log

mspensieri:
@lemieux


lemieux:
Oh wow that one survived!

:+1:

PR #131, opened 2015-11-20
Accept signedUpAt override in js widget

alavers:
@lemieux @mspensieri @dannytranlx @Mario54

### DIFF
--- a/src/js/models/appUser.js
+++ b/src/js/models/appUser.js
@@ -13,7 +13,9 @@ var AppUser = module.exports = Backbone.AssociatedModel.extend({
     },
 
     parse: function(data) {
-        return data.appUser;
+        var appUser = data.appUser;
+        appUser.signedUpAt = new Date(data.appUser.signedUpAt);
+        return appUser;
     },
 
     url: function() {
@@ -82,6 +84,7 @@ var AppUser = module.exports = Backbone.AssociatedModel.extend({
         'givenName',
         'surname',
         'email',
+        'signedUpAt',
         'properties'
     ]
 });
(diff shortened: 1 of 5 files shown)


lemieux:
:+1:

mspensieri:
While you're at it can you change line 228 in main.js

```
user.save(userInfo, {

// TO 

user.save(_.pick(userInfo, AppUser.EDITABLE_PROPERTIES), {
```

mspensieri:
:+1:

src/js/models/appUser.js
```diff
@@ -13,7 +13,9 @@ var AppUser = module.exports = Backbone.AssociatedModel.extend({
     },
 
     parse: function(data) {
-        return data.appUser;
+        var appUser = data.appUser;
+        appUser.signedUpAt = new Date(data.appUser.signedUpAt);
```
Mario54: This is necessary because the `signedUpAt` is stored as a number in the backend?

Mario54:
:+1:

src/js/models/appUser.js
lemieux: yep

alavers: to be more specific, `Date` gets serialized to a ISO 8601 string on the wire, so unless we do this we get back a string in the backbone model when we sync. I want it to be a `Date` going in and a date coming back out again.

Mario54: So, this takes care of parsing the response from the server, but what takes care of serializing `signedUpAt` when we make a request to the server? Or does the server accept a ISO 8601 string?

PR #135, opened 2015-11-20
Add support to update user attributes on login

lemieux:
Currently, we are only able to update the user on init. This makes possible to do it on login too. It also fixes a bug where some state was kept after logout and was not reset on login.

I also made the package pull the version from package.json instead of doing an awkward string replace operation on build.

@alavers @dannytranlx @jugarrit

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -14,6 +14,8 @@ var endpoint = require('./endpoint');
 var api = require('./utils/api');
 var storage = require('./utils/localStorage');
 
+var packageInfo = require('../../package.json');
+
 var SK_STORAGE = 'sk_deviceid';
 
 // appends the compile stylesheet to the HEAD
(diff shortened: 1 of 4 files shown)


mspensieri:
:+1:

src/js/main.js
```diff
@@ -116,10 +109,17 @@ _.extend(Smooch.prototype, Backbone.Events, {
 
         endpoint.sdkVersion = this.VERSION;
 
-        return this.login(options.userId, options.jwt);
+        return this.login(options.userId, options.jwt, _.pick(options, AppUser.EDITABLE_PROPERTIES));
     },
 
-    login: function(userId, jwt) {
+    login: function(userId, jwt, attributes) {
+        if (arguments.length === 2 && _.isObject(jwt)) {
```
jugarrit: Ehhh this is rather confusing to me.
Is it a big deal if we have the second param as 'null' when we only want to supply attributes?

lemieux: Yes it is for people who won't use jwt authentication.

mspensieri: @jugarrit optional arguments are awesomesauce

jugarrit:
:+1:

src/js/main.js
lemieux: ES6 makes this way easier. :)

PR #137, opened 2015-11-26
Expose message:sent and message:received events

Mario54:
@alavers @lemieux @dannytranlx @mspensieri 

Client applications can now add event handlers to `message:sent` and `message:received` events and add any behavior they desire.

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -302,11 +302,11 @@ _.extend(Smooch.prototype, Backbone.Events, {
     },
 
     _subscribeToEvents: function() {
-        this._chatController.on('message:received', _(function(message) {
+        this.listenTo(this._chatController, 'message:received', _(function(message) {
             this.trigger('message:received', message);
         }).bind(this));
 
-        this._chatController.on('message:sent', _(function(message) {
+        this.listenTo(this._chatController, 'message:sent', _(function(message) {
             this.trigger('message:sent', message);
         }).bind(this));
     },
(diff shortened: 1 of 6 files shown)


test/specs/chatController.spec.js
```diff
@@ -97,6 +97,16 @@ describe('ChatController', function() {
                 done();
             }).catch(done);
         });
+
+        it('should trigger message:sent', function(done) {
```
mspensieri: `done` callback is icky.

```
it('should trigger message:sent', function() {
    messageSentSpy = sandbox.spy();
    chatController.once('message:sent', messageSentSpy);

    return chatController.sendMessage(message).then(function() {
        messageSentSpy.should.have.been.calledOnce;
    });
});
```

src/js/main.js
```diff
@@ -298,6 +301,16 @@ _.extend(Smooch.prototype, Backbone.Events, {
         }.bind(this));
     },
 
+    _subscribeToEvents: function() {
+        this._chatController.on('message:received', _(function(message) {
```
lemieux: use `this.listenTo(this._chatController, 'message:received', ...)` instead. That way, if the chatController is destroy and reinitialize again, the old instance can be garbage collected and vice-versa, if the Smooch object is destroyed..

dannytranlx: thanks Marionette! :grin:

lemieux:
:+1:

mspensieri:
:+1:

PR #139, opened 2015-12-01
Switch to Webpack

lemieux:
This replaces browserify by webpack and also couple of other dev things.

The `example` files are no longer here because the page is now served by a dev express server that also handles serving static files/hot-loader server.

Instead of accessing `http://localhost:8282/example/index.html` you can now go directly to `http://localhost:8282/`.

Builds commands were moved to npm scripts but I aliased them  in grunt too so nothing changed.

Also, it no longer generate a non-minified js file, but it output the sourcemap on the side (which was stripped previously). I also changed the target files for both npm and bower so they use the packaged file instead of the src file. This should makes it easier for everyone to use Smooch in their project.

@alavers @jugarrit @dannytranlx


Gruntfile.js
```diff
@@ -166,8 +70,11 @@ module.exports = function(grunt) {
             js: {
                 // Files to be uploaded.
                 upload: [{
-                    src: 'dist/smooch.min.js',
+                    src: 'dist/smooch.js',
                     dest: 'smooch.min.js'
+                },{
+                    src: 'dist/smooch.js.map',
```
alavers: we're uploading source maps to s3?

lemieux: why not?

lemieux: makes it easier to debug for people who integrates it and we can get better error output from people who report errors.

alavers: Makes sense

alavers: Lol I had already started looking a week ago but then squirrel.

When would these source maps get downloaded in prod? When you open dev tools?

lemieux: yes

karma.conf.js
```diff
@@ -12,10 +19,10 @@ module.exports = function(config) {
 
         // frameworks to use
         // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
-        frameworks: ['mocha', 'browserify', 'sinon-chai', 'phantomjs-shim', 'source-map-support'],
+        frameworks: ['mocha', 'sinon-chai', 'phantomjs-shim'],
```
alavers: does webpack furnish us with source maps if we debug a test?

lemieux: see below in preprocessors, I need to debug a test to confirm though

mspensieri:
:+1:

make-webpack-config.js
```diff
@@ -0,0 +1,172 @@
...
+        smooch: './src/js/main'
+    };
+
+    var loaders = {
+        'jsx': options.hotComponents ? ['react-hot-loader', 'babel-loader'] : 'babel-loader',
+        'js': {
+            loader: 'babel-loader',
+            include: path.join(__dirname, 'src/js')
+        },
+        'json': 'json-loader',
```
alavers: Is there any way to trigger a hot load (or rebuild) whenever config.json is modified? (separate PR OK)

lemieux: I don't think so

alavers: there must be some watch equivalent for the dev server

lemieux: oh, yeah since the config is required somewhere in the code, the bundle will rebuild itself if changed but it won't trigger a hot-module replacement.. maybe a full page reload though

webpack.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = require('./make-webpack-config')({
```
alavers: is this used?

make-webpack-config.js
```diff
@@ -0,0 +1,172 @@
...
+        {
+            test: /\.js$/,
+            loader: 'imports?define=>false'
+        },
+        {
+            test: /src\/js\/main/,
+            loader: 'expose?Smooch'
+        }
+    ];
+    var alias = {
```
alavers: what are these for?

lemieux: I guess I can remove these, it was part of the boilerplate, but it's mostly the same as the alias config in requirejs.

make-webpack-config.js
```diff
@@ -0,0 +1,172 @@
...
+        filename: '[name].js' + (options.longTermCaching ? '?[chunkhash]' : ''),
+        chunkFilename: (options.devServer ? '[id].js' : '[name].js') + (options.longTermCaching ? '?[chunkhash]' : ''),
+        sourceMapFilename: '[file].map',
+        library: 'Smooch',
+        libraryTarget: 'umd',
+        umdNamedDefine: true,
+        pathinfo: options.debug
+    };
+
+    var excludeFromStats = [
```
alavers: why only exclude react and redux? shouldn't we exclude everything under node_modules?

lib/SimpleRenderer.js
```diff
@@ -0,0 +1,22 @@
+var fs = require('fs');
```
alavers: What do you think about http://blog.heroku.com/archives/2015/11/10/node-habits-2016#4-stick-with-lowercase ?

make-webpack-config.js
lemieux: You are probably right, boilerplate again.

webpack.config.js
lemieux: it's the default build you would get by running `webpack` in your console.

lib/SimpleRenderer.js
alavers: I kind of think 

> While OSX and Windows will treat 'myclass.js' and 'MyClass.js' equivalently, Linux won't.

means use upper and lower case, but never rely on case insensitive requires. Eg if we make sure to never do this: `require('simplerenderer')` then we're OK. Even better if we can lint it.

lemieux: absolutely.. boilerplate :p

package.json
```diff
@@ -39,20 +44,23 @@
...
-    "browserify": "^10.1.3",
+    "babel": "^6.1.18",
+    "babel-core": "^6.2.1",
+    "babel-loader": "^6.2.0",
+    "body-parser": "^1.14.1",
     "chai": "^2.3.0",
     "chai-as-promised": "^5.0.0",
     "chai-things": "^0.2.0",
+    "css-loader": "^0.23.0",
+    "ejs-loader": "^0.2.1",
```
alavers: Is this also from the starter project? Curious where ejs is being used.

lemieux: in all templates?

alavers:
:+1:

package.json
alavers: Derp. ofc

dannytranlx:
:+1:

karma.conf.js
lemieux: confirmed, you have the original source

lib/SimpleRenderer.js
lemieux: @alavers so what's the change to be done here? I'm kind of confused

lemieux: were you just talking about having a convention?

PR #141, opened 2015-12-03
Remove user change listener on chatView

### LINKED ISSUE #140: Undesired behavior: Smooch.updateUser opens the chat window
Invoking `Smooch.updateUser` makes chat window open up as soon as promise is resolved. Though chat window doesn't seem to be "fully" in open state, as header needs to be clicked twice to make it close itself.

To produce, go to https://dashboard.razorpay.com and run `Smooch.updateUser({givenName: 'Hello World'})` in console.

I've checked with my custom code removed, the bug is still there. Things work normally with old supportkit.min.js, so this should be recent.

lemieux:
This caused the widget to open whenever the user changed (on updateUser for instance). I'm not even sure why it was there in the first place... if we want to do something like that, we should do it only on first appMaker message or something. I'll just remove it for now.

This fixes #140.

### DIFF
--- a/src/js/views/chatView.js
+++ b/src/js/views/chatView.js
@@ -13,10 +13,6 @@ module.exports = Marionette.LayoutView.extend({
         'focus @ui.wrapper': 'focus'
     },
 
-    modelEvents: {
-        'change': 'open'
-    },
-
     ui: {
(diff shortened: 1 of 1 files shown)


mspensieri:
:+1:

PR #143, opened 2015-12-11
Service url support for dev page

lemieux:
The webpack transition made the service Url unusable in the dev environment.

This brings it back. We were replacing it directly in the bundle in dev, but now.. there's in no bundle because webpack is serving it in memory.

@dannytranlx @alavers

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,6 +10,7 @@
         /* global Smooch: false */
 
         window.SmoochOptions = {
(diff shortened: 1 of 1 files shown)


dannytranlx:
You need to add it to the template too: https://github.com/smooch/smooch-js/blob/integration/src/index.html#L20

dannytranlx:
:+1:

mspensieri:
:+1:

PR #145, opened 2016-01-04
React / Redux + better test coverage

lemieux:
This is a big chunk of changes to move away from Marionette and move towards React.
This PR includes a bigger test suite that covers almost everything in the widget.

It also leverages https://github.com/lemieux/smooch-core-js to talk with our API.

Happy reviewing @alavers @jpjoyal @dannytranlx @jugarrit @mspensieri  @Mario54 @spasiu :)


src/js/actions/auth-actions.js
```diff
@@ -0,0 +1,15 @@
...
+export const RESET_AUTH = 'RESET_AUTH';
+
+export function setAuth(props) {
+  return {
+    type: SET_AUTH,
+    props: props
+  };
+}
+
+export function resetAuth() {
```
Mario54: What do the `SET_AUTH` and `RESET_AUTH` actions do?

src/js/actions/conversation-actions.js
```diff
@@ -0,0 +1,33 @@
...
+export function setConversation(props) {
+  return {
+    type: SET_CONVERSATION,
+    conversation: props
+  }
+}
+
+export function messageAdded(props) {
+  return {
+    type: MESSAGE_ADDED,
```
Mario54: Should it be something like `ADD_MESSAGE` instead? What's the convention regarding action names?

src/js/components/chat-input.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+        this.state = {
+            text: ''
+        };
+
+        this.onChange = this.onChange.bind(this);
+        this.sendMessage = this.sendMessage.bind(this);
+    }
+
+    onChange(e) {
```
Mario54: `onTextChange`? Not clear what's changing

src/js/components/chat-input.jsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    sendMessage(e) {
+        e.preventDefault();
+        const text = this.state.text;
+        if (!!text.trim()) {
+            this.setState({
+                text: ''
+            });
+            sendMessage(text);
```
mspensieri: This is pretty confusing, looks like a recursive call. Any way to make it clearer?

src/js/actions/auth-actions.js
lemieux: To this question, I'd say : check the reducer.

In short, it removes all traces of jwt or app token from the store.

lemieux: removes/set

src/js/actions/conversation-actions.js
lemieux: yeah `ADD_MESSAGE` is probably a better name

src/js/components/chat-input.jsx
lemieux: well, the component is clearly the ChatInput, nothing else could change...

src/js/components/chat-input.jsx
lemieux: errr true, I'll change the name of the event handler

src/js/components/conversation.jsx
```diff
@@ -0,0 +1,45 @@
...
+    componentDidMount() {
+        setTimeout(this._scrollToBottom.bind(this));
+    }
+
+    componentDidUpdate() {
+        setTimeout(this._scrollToBottom.bind(this));
+    }
+
+    render() {
+        const messages = this.props.conversation.messages.map((message) => <MessageComponent key={ message._id } {...message} />);
```
mspensieri: How does string parsing work in a jsx file? i.e. how does it know that `<MessageComponent key={ message._id } {...message} />`, and the stuff in between `return ()` are strings?

src/js/components/message.jsx
```diff
@@ -0,0 +1,54 @@
...
+            <img className="sk-msg-avatar" src={ this.props.avatarUrl } />
+            );
+
+        let text = this.props.text.split('\n').map((item, index) => {
+            if (!item.trim()) {
+                return;
+            }
+
+            return (
+                <span key={ index }><span dangerouslySetInnerHTML={ createMarkup(autolink(item, {
```
mspensieri: `dangerouslySetInnerHTML` why is it dangerous?

src/js/actions/auth-actions.js
Mario54: Ok, thanks. Wasn't sure where it was used, but then realized it was to initialize the smooch core module.

src/js/components/conversation.jsx
lemieux: Good question, but that's mostly babel figuring out what's up in a jsx file. It basically transpiles it to proper JS. See https://facebook.github.io/react/#examples

src/js/services/conversation-service.js
```diff
@@ -0,0 +1,76 @@
...
+import { initFaye } from 'utils/faye';
+
+export function sendMessage(text) {
+    var sendFn = () => {
+        const message = {
+            text: text,
+            role: 'appUser'
+        };
+
+        store.dispatch(messageAdded(message));
```
Mario54: So, right now we add the message before making the request to the server? What happens if the request to add a message to the conversation is rejected?

src/js/reducers/app-state-reducer.js
```diff
@@ -0,0 +1,75 @@
...
+    serverURL: 'https://api.smooch.io/'
+};
+
+export function AppStateReducer(state = INITIAL_STATE, action) {
+    switch (action.type) {
+        case RESET:
+            return Object.assign({}, INITIAL_STATE);
+
+        case ENABLE_SETTINGS:
+            return Object.assign({}, state, {
```
mspensieri: Such duplication here. Is this easier to read?

```
function getStateChange(action) {
    switch (action.type) {
        case ENABLE_SETTINGS:
            return {
                settingsEnabled: true
            };

        case DISABLE_SETTINGS:
    /*..... */
    }
}

export function AppStateReducer(state = INITIAL_STATE, action) {
    if (action.type === RESET) {
        return Object.assign({}, INITIAL_STATE);
    }

    return Object.assign({}, state, getStateChange(action));
}
```

src/js/components/message.jsx
Mario54: https://facebook.github.io/react/tips/dangerously-set-inner-html.html

lemieux: because it is dangerous to use raw html

src/js/services/conversation-service.js
lemieux: Currently nothing, but that was the behavior before. We should definitely do something globally about that sort of thing in the `catch` of these promises. 

If we add it only on server response, it creates a unacceptable delay for the user.

src/js/components/widget.jsx
```diff
@@ -0,0 +1,36 @@
...
+import { Notification } from 'components/notification.jsx';
+import { ChatInput } from 'components/chat-input.jsx';
+
+export class WidgetComponent extends Component {
+
+    render() {
+        const mainComponent = this.props.appState.settingsVisible ? <Settings /> : <Conversation />;
+        const footer = this.props.appState.settingsVisible ? null : <ChatInput />;
+
+        let className = typeof this.props.appState.widgetOpened === 'undefined' ? '' :
```
Mario54: Why check the `typeof` instead of checking if it's falsy?

src/js/reducers/app-state-reducer.js
lemieux: I agree there is duplication, but IMO it's easier to read than splitting the logic in two different places instead of having everything in the same switch statement.

src/js/components/widget.jsx
lemieux: That's a good question and it's mainly due to the animation to open and close the widget. `undefined` means it's in its default state, was never opened, was never closed. Once opened, the widget add CSS classes to manage the animations to open and close.

I should probably put that explanation in a comment on top of that line.

src/js/services/conversation-service.js
Mario54: So we would optimistically create the message and then if it's rejected by the server, we'd show an error message?

mspensieri:
Looks pretty nice :)

src/js/services/conversation-service.js
lemieux: Yes, this is the behavior of a lot of apps (but they usually show a small `Sending` on the side of the message). On error, you usually get a `Resend` or something like that.

mspensieri: Sounds like a feature request. In iOS (and Android I think) it says "sending..." and then "message not delivered" if it fails, or sending goes away when it succeeds

lemieux: I would replicate the same behavior on the web.

lemieux: But yeah... this is a feature request in a separate PR.

src/js/reducers/app-state-reducer.js
Mario54: I'd say keep the duplication too.

src/js/components/widget.jsx
Mario54: Ah, I see.

src/js/components/conversation.jsx
```diff
@@ -0,0 +1,45 @@
...
+                { messages }
+                <div className="sk-logo" ref="logo">
+                    <a href="https://smooch.io/?utm_source=widget" target="_blank">In-App Messaging by <img className="sk-image" src={ require('images/logo_webwidget.png') } alt="Smooch" /> <img className="sk-image-retina" src={ require('images/logo_webwidget_2x.png') } alt="Smooch" /></a>
+                </div>
+            </div>
+            );
+    }
+}
+
+export const Conversation = connect((state) => {
```
Mario54: Let me know if this is wrong: The `connect` function specifies how the state of the store is mapped to the props of the component?

lemieux: That and it let the component know when the store changed so it can rerender.

src/js/actions/conversation-actions.js
```diff
@@ -0,0 +1,33 @@
...
+    message: Object.assign({
+      _id: Math.random(),
```
Mario54: Is it okay to use `Math.random()` for this? It'll only return numbers in the [0,1) range.

It's only used for the `key` prop of the `MessageComponent` for now, does React expect the `key` to be an integer?

test/specs/components/chat-input.spec.jsx
```diff
@@ -0,0 +1,121 @@
...
+            };
+            component.sendMessage(event);
+
+            event.preventDefault.should.have.been.calledOnce;
+        });
+
+        for (let value of ['', '      ']) {
+            it('should do nothing if the current state is blank', () => {
+                component.state.text = value;
+                component.sendMessage({
```
Mario54: Shouldn't you simulate a click of the send button instead of doing this manually?

src/js/actions/conversation-actions.js
lemieux: It's used for the key prop and it doesn't need to be an integer.

test/specs/components/chat-input.spec.jsx
lemieux: Look at 2 tests above, it's simulating the click and make sure the handler is called.

This test ensures the handler works as expected.

mspensieri:
:+1: I think? This deserves a bug bash

package.json
```diff
@@ -23,20 +26,20 @@
...
+    "babel-preset-es2015": "^6.1.2",
+    "babel-preset-react": "^6.1.2",
+    "babel-preset-stage-2": "^6.1.2",
+    "babel-runtime": "^6.0.14",
+    "deep-equal": "^1.0.1",
     "faye": "^1.1.2",
-    "jquery": "^1.11.3",
-    "lodash.bindall": "^3.1.0",
-    "marionette.behaviors": "0.0.13",
-    "underscore": "1.8.3",
```
alavers: karma.conf.js still uses underscore

src/index.html
```diff
@@ -20,6 +20,10 @@
             properties: PROPERTIES,
             emailCaptureEnabled: true
         };
+        
+        Smooch.on('message', function(message) {
+          console.log('Message received', message);
```
alavers: remove?

lemieux: I was thinking about leaving it there since it's for debugging purpose and this whole page is for debugging purpose. Shows that it still works.

package.json
lemieux: Added it back in dev deps.

src/js/components/chat-input.jsx
```diff
@@ -0,0 +1,66 @@
...
+        if (getReadTimestamp() > 0) {
+            updateReadTimestamp(0);
+        }
+    }
+
+    render() {
+        return (
+            <div id="sk-footer">
+                <form onSubmit={ this.onSendMessage }>
+                    <input ref="input"
```
alavers: is this `ref` being used?

mspensieri:
Bug Reports:

* it doesn't find `config/default/config` anymore
* `serviceUrl` param in config is not respected, always calls `api.smooch.io`
* Unread badge number never increments from 1
* Read messages doesn't persist between page loads (see gif, when the widget disappears is when I refreshed the page)
unread_badge

alavers:
why single quotes in JS but double quotes in JSX? I noticed react samples do htis but I was wondering if there was a reason.

src/js/components/chat-input.jsx
lemieux: in tests I think

lemieux:
@alavers I dunno.. looks more html-like maybe?

dannytranlx:
Tagline not positioned correctly?

Chrome OSX El Capitan

image

@mspensieri Odd, I wasn't able to reprod your issue.

lemieux:
default config support added

src/js/components/header.jsx
```diff
@@ -0,0 +1,97 @@
...
+        let unreadTimestamp = this.props.appState.messageReadTimestamp;
+
+        let unreadMessagesCount = unreadTimestamp <= 0 ? 0 : this.props.conversation.messages.reduce((count, message) => {
+            if (message.role !== 'appUser' && unreadTimestamp <= message.received) {
+                return count + 1;
+            }
+            return count;
+        }, 0);
+
+        let unreadBadge = !settingsVisible && unreadMessagesCount > 0 ? (
```
alavers: style nit, don't need to wrap all jsx in (). Eg this could be:

```
        let unreadBadge = !settingsVisible && unreadMessagesCount > 0 ?
            <div id="sk-badge">
                { unreadMessagesCount }
            </div> : null;
```

lemieux: we could, but I'd prefer to either do it nowhere.. or everywhere.. and since it needs to be like that with `return` ....

src/js/reducers/app-state-reducer.js
```diff
@@ -0,0 +1,80 @@
+import { ENABLE_SETTINGS, DISABLE_SETTINGS, TOGGLE_WIDGET, OPEN_WIDGET, CLOSE_WIDGET, SHOW_SETTINGS, HIDE_SETTINGS, SHOW_SETTINGS_NOTIFICATION, HIDE_SETTINGS_NOTIFICATION, SET_SERVER_URL, SET_EMAIL_READONLY, UNSET_EMAIL_READONLY, UPDATE_READ_TIMESTAMP } from 'actions/app-state-actions';
```
alavers: ew. Is there no better way?

lemieux: `import * as AppStateActions from 'actions/app-state-actions'` and do `AppStateActions.UNSET_EMAIL_READONLY` everywhere.. the formatter won't let me put that on mutliple lines

alavers: why do action consts need to be in separate files?
what if we had a single `actions.js` one to rule them all
then you could `import * as action from 'actions'`

lemieux: it's not just the constants in those files, there are also the action creators functions

src/js/stores/middlewares/messages.js
```diff
@@ -0,0 +1,24 @@
+import { ADD_MESSAGE } from 'actions/conversation-actions';
+import { showSettingsNotification } from 'actions/app-state-actions';
+
+
+export function firstMessage(store) {
+    return function(next) {
+        return function(action) {
+            let result = next(action);
+            let state = store.getState();
+            if (state.appState.settingsEnabled && action.type === ADD_MESSAGE && action.message.role === 'appUser' && !state.user.email) {
```
alavers: Seems odd that we have to handle first app user messages with a middleware.

Is there a way we could dispatch a `FIRST_USER_MESSAGE` event? Maybe do it from `conversation-service`? Then it's up to the settings view to listen for this event and decide whether or not to fire.

src/js/services/auth-service.js
```diff
@@ -0,0 +1,11 @@
+import { store } from 'stores/app-store';
+import { core } from 'services/core';
+
+import { setUser } from 'actions/user-actions';
+
+export function login(props) {
+    return core().appUsers.init(props).catch((e) => {
+        console.log(e); //eslint-disable-line no-console
```
alavers: I added lint exceptions to these console.logs but should I have? Do these errors belong in the console?

src/js/stores/middlewares/messages.js
alavers: I guess the reducer would need to do something with this `FIRST_USER_MESSAGE ` event, not the view itself. Does it make sense to check state and dispatch `showSettingsNotification()` from a reducer?

alavers: I guess not lol https://github.com/rackt/redux/issues/368

src/js/services/auth-service.js
lemieux: It's pretty much up to use.. I don't want people to complain because it fails silently.. and since everything is handled in promises now, you have to catch it to know what failed.

src/js/stores/middlewares/messages.js
lemieux: I was once the fence about this one really... either a middleware or in the service directly. I think the service is better now.. I think it's better to run this code everytime the user sends a message than everytime an action is dispatched.

src/js/services/auth-service.js
mspensieri: Don't ajax errors get logged to the console anyways?

lemieux: I was about to say the ajax error is not really descriptive... but my console.log is not either...

image

lemieux: And unhandled rejections are logged... so... that console.log is going away.

src/js/stores/middlewares/messages.js
lemieux: Moved to the conversation service

src/js/utils/events.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    off(event, handler) {
+        let map = this[listeners];
+        if (map.has(event)) {
+            map.get(event).delete(handler);
+        }
+    }
+
+    trigger(event, args) {
```
alavers: I think perhaps you want:

```
trigger(event, ...args) {
```

and

```
map.get(event).forEach(handler => handler(...args));
```

src/js/utils/functions.js
```diff
@@ -0,0 +1,8 @@
+export function pick(o, fields) {
```
alavers: Instead of setting this precedent what do you think about https://lodash.com/custom-builds ?

(very much out of scope of this PR but I'm curious to know what you think)

lemieux: I don't why I didn't do this.. lodash is publishing each of its functions separately https://www.npmjs.com/package/lodash.pick

src/js/utils/redux.js
```diff
@@ -0,0 +1,6 @@
+import createLogger from 'redux-logger';
```
alavers: Is this file used anywhere?

src/js/utils/events.js
lemieux: maybe it shouldn't be called args.. I thought about doing that, but the current eventing system (backbone events) use `(event, options)`

src/js/utils/redux.js
lemieux: it's gone now, I figured that out too :)

src/stylesheets/notifications.less
```diff
@@ -1,42 +1,62 @@
-.sk-notifications {
+.sk-notification-container {
+    box-shadow: 0 1px 3px rgba(0, 0, 0, .1);
     position: absolute;
     z-index: 1;
 
-    overflow: hidden;
+    .sk-notifications {
```
alavers: nit, this file ought to be called `notification.less` so it matches the transitionName.
And this selector should be `.sk-notifications`

lemieux: true

src/js/utils/functions.js
lemieux: fixed with lodash.pick

test/specs/components/chat-input.spec.jsx
```diff
@@ -0,0 +1,121 @@
+import sinon from 'sinon';
```
alavers: Hmm. Tests seem way nicer with react... :)

test/specs/components/conversation.spec.jsx
```diff
@@ -0,0 +1,64 @@
...
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    it('should use the ui text for the intro', () => {
+        component.refs.intro.textContent.should.eq(props.ui.text.introText);
+    });
+
+    it('should generate all messages in the props', () => {
+        TestUtils.scryRenderedDOMComponentsWithClass(component, 'mockedMessage').length.should.eq(props.conversation.messages.length);
```
alavers: My best guess is that some DnD nerd use the term `scry` instead of `find` so that this method looked different from `findRenderedDOMComponentWithClass` (singular). Is that right?

lemieux: I think find exists, but returns only one element, scry returns an array.

lemieux: and find crashes if not found... or something like that

test/utils/react.js
```diff
@@ -0,0 +1,32 @@
...
+    });
+}
+
+export function scryRenderedDOMComponentsWithAttribute(tree, attr, value = '') {
+    return TestUtils.findAllInRenderedTree(tree, function(inst) {
+        return TestUtils.isDOMComponent(inst) && inst.getAttribute(attr) === value;
+    });
+}
+
+export function findRenderedDOMComponentsWithId(tree, id) {
```
alavers: Doesn't this already exist and isn't it called `TestUtils.findRenderedDOMComponentWithId` ?

lemieux: does it? lol

src/stylesheets/notifications.less
lemieux: fixed

test/specs/components/header.spec.jsx
```diff
@@ -0,0 +1,220 @@
...
+            headerText: 'Header',
+            settingsHeaderText: 'Settings'
+        }
+    }
+};
+
+describe('Header', () => {
+    afterEach(() => {
+        sandbox.restore();
+        defaultProps.actions.showSettings = sandbox.spy();
```
alavers: What if we change `const defaultProps` to `let props` and initialize it in a `beforeEach` like this:

```
    beforeEach( () => {
        props = {
            ...
        };
    });
```

You wouldn't need to reset the spies in this afterEach and it would reduce the possibility of tests interfering with another.

test/utils/react.js
lemieux: oh, no it doesn't... there's findRenderedDOMComponentsWithTag and findRenderedDOMComponentsWithClass, but no findRenderedDOMComponentsWithId .. that's why I made one... I guess React doesn't expect us to use ids...

test/specs/components/header.spec.jsx
lemieux: fixed

test/utils/react.js
alavers: Ah :+1:

lemieux:
@dannytranlx tag line position fixed

alavers:
Press CTRL + D to Bookmark this page and get quick access to your favorite ASCII Art.
Thumbs Up Collection ASCII Text Art











╱╱┏╮
╱╱┃┃
▉━╯┗━╮
▉┈┈┈┈┃
▉╮┈┈┈┃
╱╰━━━╯







▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓███▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓█▓▓██▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓█▒▒███▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓███▒▒█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓███▒▒█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓██▒░▒█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓█▓░▒▒█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓█▓▒▒▒█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓█▓▒▒▒██▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓█▓▒▒▒▒██▓▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
▒▓█▓▒▒▒▒▒██████████████████████▓▒▒▒
▒▓█▓▒▒▒▒▒▒▒▓█████████████████████▒▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒▒▒░░░░░░░░░░░░░. ▓█▓▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒▒▒▓▓▓▓▓▓▓▓▓▓▓▓▓███▒▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒▓█████████████████▒▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒██░░░░░░░░░░░░░░▓█▓▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒▓█▓▓▓▓▓▓▓▓▓▓▓▓▓▓██▒▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒▒█████████████████▒▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒██░░░░░░░░░░░░░░▓█▓▒
▒▓█▓▒▒▒▒▒▒▒▒▒▒▒▓█▓▒▓▓▓▓▓▓▓▓▓▓▓▓██▒▒
▒▒██▒▒▒▒▒▒▒▒▒▒▒▒█████████████████▒▒
▒▒▓██▒▒▒▒▒▒▒▒▒▒██▒░▒▒▒▒▒▒▒▒▒▒░░▓█▓▒
▒▒▒▓███▓▓▓▓▓▓▓▒██▓▒▓▓▓▓▓▓▓▓▓▒▒▒██▓▒
▒▒▒▒▒▓██████████████████████████▒▒▒













────────────────────░███░
───────────────────░█░░░█░
──────────────────░█░░░░░█░
─────────────────░█░░░░░█░
──────────░░░───░█░░░░░░█░
─────────░███░──░█░░░░░█░
───────░██░░░██░█░░░░░█░
──────░█░░█░░░░██░░░░░█░
────░██░░█░░░░░░█░░░░█░
───░█░░░█░░░░░░░██░░░█░
──░█░░░░█░░░░░░░░█░░░█░
──░█░░░░░█░░░░░░░░█░░░█░
──░█░░█░░░█░░░░░░░░█░░█░
─░█░░░█░░░░██░░░░░░█░░█░
─░█░░░░█░░░░░██░░░█░░░█░
─░█░█░░░█░░░░░░███░░░░█░
░█░░░█░░░██░░░░░█░░░░░█░
░█░░░░█░░░░█████░░░░░█░
░█░░░░░█░░░░░░░█░░░░░█░
░█░█░░░░██░░░░█░░░░░█░
─░█░█░░░░░████░░░░██░
─░█░░█░░░░░░░█░░██░█░
──░█░░██░░░██░░█░░░█░
───░██░░███░░██░█░░█░
────░██░░░███░░░█░░░█░
──────░███░░░░░░█░░░█░
──────░█░░░░░░░░█░░░█░
──────░█░░░░░░░░░░░░█░
──────░█░░░░░░░░░░░░░█░
──────░█░░░░░░░░░░░░░█░





★░░░░░░░░░░░████░░░░░░░░░░░░░░░░░░░░★ ★░░░░░░░░░███░██░░░░░░░░░░░░░░░░░░░░★ ★░░░░░░░░░██░░░█░░░░░░░░░░░░░░░░░░░░★ ★░░░░░░░░░██░░░██░░░░░░░░░░░░░░░░░░░★ ★░░░░░░░░░░██░░░███░░░░░░░░░░░░░░░░░★ ★░░░░░░░░░░░██░░░░██░░░░░░░░░░░░░░░░★ ★░░░░░░░░░░░██░░░░░███░░░░░░░░░░░░░░★ ★░░░░░░░░░░░░██░░░░░░██░░░░░░░░░░░░░★ ★░░░░░░░███████░░░░░░░██░░░░░░░░░░░░★ ★░░░░█████░░░░░░░░░░░░░░███░██░░░░░░★ ★░░░██░░░░░████░░░░░░░░░░██████░░░░░★ ★░░░██░░████░░███░░░░░░░░░░░░░██░░░░★ ★░░░██░░░░░░░░███░░░░░░░░░░░░░██░░░░★ ★░░░░██████████░███░░░░░░░░░░░██░░░░★ ★░░░░██░░░░░░░░████░░░░░░░░░░░██░░░░★ ★░░░░███████████░░██░░░░░░░░░░██░░░░★ ★░░░░░░██░░░░░░░████░░░░░██████░░░░░★ ★░░░░░░██████████░██░░░░███░██░░░░░░★ ★░░░░░░░░░██░░░░░████░███░░░░░░░░░░░★ ★░░░░░░░░░█████████████░░░░░░░░░░░░░★ ★░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░★







┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌┌█████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌███████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌███┌┌┌██┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌██┌┌┌┌██┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌███┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌███┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌██┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌███┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌███┌┌┌┌┌████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌┌██┌┌┌┌┌┌████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌┌███┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌┌┌███┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌┌┌┌██┌┌┌┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌███┌███┌┌┌┌┌┌┌┌┌┌██┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌████████████┌┌┌┌┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌
┌┌████████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌┌
┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌█████████┌┌
███┌┌┌┌█████████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌█████┌
██┌┌┌███████┌████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌███┌
██┌┌┌┌███┌┌┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌██┌
███┌┌┌┌┌┌┌┌┌┌┌█████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌██┌
┌███┌┌┌┌┌┌┌████████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌██┌
┌┌████████████┌┌┌████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌██┌
┌███┌██████┌┌┌┌┌┌┌████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌██┌
┌███┌┌┌┌┌┌┌┌┌┌┌██████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌██┌
┌┌████┌████┌██████████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌██┌
┌┌┌████████████┌┌┌┌┌███┌┌┌┌┌┌┌┌┌┌┌┌┌███┌
┌┌┌┌██┌┌┌┌┌┌┌┌┌┌┌███████┌┌┌┌┌┌┌███████┌┌
┌┌┌┌████┌┌┌┌┌┌████████┌┌┌┌┌┌┌┌████████┌┌
┌┌┌┌┌████████████┌┌┌███┌┌┌┌┌███┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌███┌█┌█┌┌┌┌┌┌███┌┌┌███┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌███┌┌┌┌┌┌█████┌┌█████┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌██████████████████┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌██████████████┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌
┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌┌

spasiu:
awesome. and most of the tests look so much tighter then with backbone

PR #146, opened 2016-01-11
Support for buy buttons

lemieux:
image

This adds support for Stripe on buy buttons. 

It's currently missing error handling when we call our backend, but it should come in shortly in this PR or a second one. It's also missing tests.

I'd like to get your feedback on this :)

@jpjoyal @dannytranlx @Mario54 @alavers @mspensieri

### DIFF
--- a/src/js/components/action.jsx
+++ b/src/js/components/action.jsx
@@ -44,8 +44,6 @@ export class ActionComponent extends Component {
         this.setState({
             state: 'processing'
         });
-
-        //  this.refs.stripe.onClick();
     }
 
     render() {
@@ -58,7 +56,6 @@ export class ActionComponent extends Component {
                     <StripeCheckout token={ this.onStripeToken.bind(this) }
                                     stripeKey={ publicKeys.stripe }
                                     email={ user.email }
-                                    allowRememberMe={ false }
                                     amount={ this.props.amount }
                                     currency={ this.props.currency.toUpperCase() }>
                         <button className='btn btn-sk-primary' onClick={ this.onStripeClick.bind(this) }>
--- a/src/js/components/action.jsx
+++ b/src/js/components/action.jsx
@@ -1,0 +1,90 @@
+import React, { Component } from 'react';
+import StripeCheckout from 'react-stripe-checkout';
+
+import { store } from 'stores/app-store';
+import { createTransaction } from 'services/stripe-service';
+import { immediateUpdate } from 'services/user-service';
+
+import { LoadingComponent } from 'components/loading';
+
+export class ActionComponent extends Component {
+    constructor(...args) {
+        super(...args);
+
+        this.state = {
+            state: this.props.state
+        };
+    }
+
+    onStripeToken(token) {
+        let user = store.getState().user;
+        let promises = [];
+        if (!user.email) {
+            promises.push(immediateUpdate({
+                email: token.email
+            }));
+        }
+
+        let transactionPromise = createTransaction(this.props._id, token.id).then(() => {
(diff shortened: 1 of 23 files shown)


src/js/components/action.jsx
```diff
@@ -0,0 +1,90 @@
...
+        if (this.props.type === 'buy') {
+            let user = store.getState().user;
+            let publicKeys = store.getState().app.publicKeys;
+            let actionState = this.state.state;
+            if (actionState === 'offered') {
+                return (
+                    <StripeCheckout token={ this.onStripeToken.bind(this) }
+                                    stripeKey={ publicKeys.stripe }
+                                    email={ user.email }
+                                    allowRememberMe={ false }
```
mspensieri: Why not?

src/js/components/action.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+        return Promise.all(promises);
+    }
+
+    onStripeClick() {
+        this.setState({
+            state: 'processing'
+        });
+
+        //  this.refs.stripe.onClick();
```
mspensieri: Uncomment or remove?

dannytranlx: ?

src/js/components/action.jsx
```diff
@@ -0,0 +1,90 @@
+import React, { Component } from 'react';
+import StripeCheckout from 'react-stripe-checkout';
```
mspensieri: How big is this lib? Will it impact load time?

src/js/components/action.jsx
lemieux: whoops.. go away

src/js/components/action.jsx
lemieux: good question...

src/stylesheets/main.less
```diff
@@ -23,7 +24,7 @@
         border-radius: 10px 10px 0 0;
         text-rendering: optimizeLegibility;
         color: #333;
-        z-index: 2147483647;
```
dannytranlx: :joy:

dannytranlx:
Man, code's pretty clean!

:+1:

src/js/components/action.jsx
lemieux: 9kb unminified... best thing is, it loads checkout only when needed... no buy button? no stripe!

It won't impact load time at all.

src/js/components/action.jsx
lemieux: added it back

mspensieri:
:+1: :+1: :+1:

src/js/components/action.jsx
```diff
@@ -0,0 +1,90 @@
...
+        let user = store.getState().user;
+        let promises = [];
+        if (!user.email) {
+            promises.push(immediateUpdate({
+                email: token.email
+            }));
+        }
+
+        let transactionPromise = createTransaction(this.props._id, token.id).then(() => {
+            this.setState({
```
mspensieri: Does this get translated into some change in the store? If the component needs to be redrawn and loses state, can it regain this state from `store.conversation.messages.actions`, or wherever it's persisted?

lemieux: this part is just local to the component. it will stay there until it's destroyed (I think we never destroy the conversation component, so unless you log out, the component will stay there). 

if the conversation is refetched, this component should retain it's state.. I should probably add something if it's refetched and the new state is `paid` it should take the paid state instead of whatever it is at that moment.

alavers:
Slowpoke :+1:

PR #147, opened 2016-01-12
Stripe app name and icon

lemieux:
Here you go, this makes use of the new endpoint and display the name and icon of the Stripe app.

@alavers @dannytranlx @Mario54

### DIFF
--- a/src/js/components/action.jsx
+++ b/src/js/components/action.jsx
@@ -12,11 +12,16 @@ export class ActionComponent extends Component {
         super(...args);
 
         this.state = {
-            state: this.props.state
+            state: this.props.state,
+            hasToken: false
         };
     }
 
     onStripeToken(token) {
+        this.setState({
+            hasToken: true
+        });
+
         let user = store.getState().user;
         let promises = [];
         if (!user.email) {
@@ -24,9 +29,6 @@ export class ActionComponent extends Component {
                 email: token.email
             }));
         }
-        this.setState({
-            state: 'processing'
-        });
 
         let transactionPromise = createTransaction(this.props._id, token.id).then(() => {
             this.setState({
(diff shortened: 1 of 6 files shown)


src/js/smooch.jsx
```diff
@@ -146,11 +147,17 @@ export class Smooch extends Observable {
         }).then((loginResponse) => {
             store.dispatch(setUser(loginResponse.appUser));
             store.dispatch(updateReadTimestamp(getReadTimestamp()));
-            
+
             if (loginResponse.publicKeys) {
                 store.dispatch(setPublicKeys(loginResponse.publicKeys));
-            }
 
+                if (loginResponse.publicKeys.stripe) {
```
Mario54: We should probably find a better way to indicate that stripe is configured for the app in the `init` response

Mario54:
:+1:

src/js/smooch.jsx
lemieux: My guess is that it's the best way... don't have the public key? no stripe for you

mspensieri: The alternative would be to have an extra key in the response, `app.stripeEnabled = !!publicKeys.stripe`, which formalizes the assumption that public key means stripe enabled.

lemieux: yeah we could do that too

lemieux: since we already have `gcmEnabled`, that could be similar

lemieux:
ehhh I added the spinner back in this PR by error.... just review too plez :) @Mario54

src/js/components/action.jsx
```diff
@@ -57,7 +76,10 @@ export class ActionComponent extends Component {
                                     stripeKey={ publicKeys.stripe }
                                     email={ user.email }
                                     amount={ this.props.amount }
-                                    currency={ this.props.currency.toUpperCase() }>
+                                    currency={ this.props.currency.toUpperCase() }
+                                    name={ stripeAccount.appName }
+                                    image={ stripeAccount.iconUrl }
+                                    closed={ this.onStripeClose.bind(this) }>
```
alavers: Is there any preference for doing this vs `this.onStripeClose = this.onStripeClose.bind(this)` in the constructor?

src/js/smooch.jsx
Mario54: Yeah having a `stripeEnabled` key sounds better

src/js/components/action.jsx
lemieux: in this case the constructor is already define but I usually prefer this to not have to add a constructor

alavers:
:+1:

Mario54:
:+1:

src/js/smooch.jsx
lemieux: I'll leave it like that for now and add a card about adding that to the backend

PR #149, opened 2016-01-13
Escape HTML before rendering messages

lemieux:
Html wasn't escaped allowing the message to be messed up if html was in the content.


lemieux:
@mspensieri @alavers

mspensieri:
:+1:

alavers:
:+1:

PR #150, opened 2016-01-13
Handle Stripe Account error

lemieux:
This should make the widget work if we have the public keys but the call to /stripe/accounts fails for some reason. The end result would be that the Checkout modal would show up with no name or icon.

@mspensieri


mspensieri:
:+1:

PR #151, opened 2016-01-13
Non-persistent and non-increasing badge count fix

lemieux:
Whenever the conversation was updated and the timestamp was at 0, it always took the last message as the unread timestamp. On startup, the ts was at 0 so it was set to the last message `received` props after fetching the conversation. I normalized a bit how we check for a lastRead update.

Also, every message coming through faye was setting the unread ts to that message ts which ended up having a count of 1 all the time.

@alavers @mspensieri

### DIFF
--- a/src/js/services/user-service.js
+++ b/src/js/services/user-service.js
@@ -2,7 +2,7 @@ import deepEqual from 'deep-equal';
 import { store } from 'stores/app-store';
 import { core } from 'services/core';
 import { setUser } from 'actions/user-actions';
-import { getConversation, connectFaye } from 'services/conversation-service';
+import { handleConversationUpdated } from 'services/conversation-service';
 
 let waitForSave = false;
 const waitDelay = 5000; // ms
@@ -57,11 +57,9 @@ export function trackEvent(eventName, userProps) {
     const user = store.getState().user;
     return core().appUsers.trackEvent(user._id, eventName, userProps).then((response) => {
         if (response.conversationUpdated) {
-            return getConversation()
-                .then(connectFaye)
-                .then(() => {
-                    return response;
-                });
+            return handleConversationUpdated().then(() => {
+                return response;
+            });
         }
 
         return response;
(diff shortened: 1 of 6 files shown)


src/js/services/user-service.js
```diff
@@ -57,11 +57,9 @@ export function trackEvent(eventName, userProps) {
     const user = store.getState().user;
     return core().appUsers.trackEvent(user._id, eventName, userProps).then((response) => {
         if (response.conversationUpdated) {
-            return getConversation()
-                .then(connectFaye)
-                .then(() => {
-                    return response;
-                });
+            return handleConversationUpdated().then(() => {
```
mspensieri: Shouldn't this only fetch the conversation if faye is not already connected?

src/js/smooch.jsx
```diff
@@ -204,10 +204,7 @@ export class Smooch extends Observable {
     updateUser(props) {
         return updateUser(props).then((response) => {
             if (response.appUser.conversationStarted) {
-                return getConversation().then((conversationResponse) => {
-                    store.dispatch(setConversation(conversationResponse.conversation));
-                    return connectFaye();
-                }).then(() => {
+                return handleConversationUpdated().then(() => {
```
mspensieri: Same question

src/js/services/user-service.js
lemieux: `connectFaye` takes care of checking if there's an existing subscription or not and either subscribe or do nothing.

mspensieri: But the `GET /conversation` call can be avoided if faye is connected, because the message will come through websockets anyways

lemieux: true, I'll wrap that part under an `if !faye subscription` block

src/js/smooch.jsx
lemieux: same here

src/js/services/user-service.js
lemieux: I put the check in `handleConversationUpdated` directly

mspensieri:
:+1:

PR #156, opened 2016-01-14
Fix dismiss notification mechanic

lemieux:
I also fixed some rendering issue for the buy button when they are in processing mode.

Will merge right away.

### DIFF
--- a/src/js/components/error-notification.jsx
+++ b/src/js/components/error-notification.jsx
@@ -10,10 +10,10 @@ import { hideErrorNotification } from 'actions/app-state-actions';
 export class ErrorNotificationComponent extends Component {
     render() {
         return (
-            <div key="content" className="sk-notification sk-notification-error">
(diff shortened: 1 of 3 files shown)


src/js/components/error-notification.jsx
```diff
@@ -10,10 +10,10 @@ import { hideErrorNotification } from 'actions/app-state-actions';
 export class ErrorNotificationComponent extends Component {
     render() {
         return (
-            <div key="content" className="sk-notification sk-notification-error">
+            <div key="content" className="sk-notification sk-notification-error" onClick={ this.props.actions.hideErrorNotification }>
                 <p>
                     <span ref="text" dangerouslySetInnerHTML={ createMarkup(this.props.message) }></span>
-                    <a href="#" className="sk-notification-close" onClick={ this.props.actions.hideErrorNotification }>&times;</a>
+                    <a href="#" className="sk-notification-close">&times;</a>
```
dannytranlx: Should we even keep the `x` in that case? Or it isn't obvious enough that you can dismiss by clicking on it?

PR #160, opened 2016-01-19
Prevent double insertion of babel-polyfill

lemieux:
This should prevent problems with site that already have it.

This would need to be tests on IE at least since that the browser that required the most it in the first place.

@mspensieri @dannytranlx

### DIFF
--- a/src/js/utils/polyfills.js
+++ b/src/js/utils/polyfills.js
@@ -1,3 +1,7 @@
+if (!global._babelPolyfill) {
+    require('babel-polyfill');
+}
+
 (function() {
     // setPrototypeOf Polyfill for =< IE10
 
--- a/make-webpack-config.js
+++ b/make-webpack-config.js
@@ -11,7 +11,7 @@ module.exports = function(options) {
     var LICENSE = fs.readFileSync('LICENSE', 'utf8');
 
     var entry = {
-        smooch: ['babel-polyfill', './src/js/utils/polyfills', './src/js/main']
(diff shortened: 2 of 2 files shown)


src/js/utils/polyfills.js
```diff
@@ -1,3 +1,7 @@
+if (!global._babelPolyfill) {
```
jpjoyal: This is evaluated a built time no ?

lemieux: Should be evaluated at runtime... but that's mainly why I want to test it in IE.. just to make sure.

barthoda:
Looking forward to a fix as we removed smooch today due to this problem

mspensieri:
:+1:

lemieux:
@jpjoyal I just tested this in IE10 and it works correctly.

I basically added `<script src="https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/6.3.14/polyfill.min.js"></script>` in the page before our script to see if the error was thrown, and it was (on integration). Then I used this branch and the error goes away.

This fix works when we load after an app that already uses babel-polyfill, but we can't do anything if we are the first to load and they don't do this check (which they have no real reason to do...).

I suggest we do this for now, but we also look into providing a polyfill-less build too for people who already has it or do not need to support old browsers. It would also result in a smaller build (it removes ~30kb).

jpjoyal:
Let's go with this. I really hesitant to start having a separate build & publication for this. Seems more trouble than worth.

jpjoyal:
:+1:

PR #162, opened 2016-01-19
Stripe tests

lemieux:
I added tests for the action component and the stripe service.

@Mario54 @mspensieri @alavers


test/specs/components/action.spec.jsx
```diff
@@ -0,0 +1,541 @@
...
+        sandbox.stub(userService, 'immediateUpdate');
+        userService.immediateUpdate.resolves();
+    });
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    after(() => {
+        Object.defineProperty(AppStore, 'store', {
```
alavers: Why is this necessary?

test/specs/components/action.spec.jsx
```diff
@@ -0,0 +1,541 @@
...
+            text: 'action text',
+            uri: 'action uri'
+        };
+
+        beforeEach(() => {
+            component = TestUtils.renderIntoDocument(<ActionComponent {...props} />);
+            componentNode = ReactDOM.findDOMNode(component);
+        });
+
+        it('should render a link with target blank', () => {
```
alavers: `beforeEach` and `it` could be merged. Do you plan to add more `it`s here?

test/specs/components/action.spec.jsx
```diff
@@ -0,0 +1,541 @@
...
+            link.target = '_self';
+        });
+    });
+
+    describe('buy action with stripe keys and offered state', () => {
+        const props = {
+            type: 'buy',
+            text: 'action text',
+            uri: 'fallback uri',
+            currency: 'USD',
```
alavers: nit, stripe lower cases currency and we should probably follow suit

lemieux: react-stripe-checkout expects it to be uppercased.

test/specs/components/action.spec.jsx
lemieux: probably not, but I've been doing that in the whole test suite..

test/specs/components/action.spec.jsx
lemieux: since `store` is not a method of `AppStore` it cannot be stubbed. The mock store above takes care of replacing the value by a mock. This part restores it at the initial value.

I really need to try babel-rewire again.

test/specs/components/action.spec.jsx
```diff
@@ -0,0 +1,541 @@
...
+            state: 'offered'
+        };
+
+        beforeEach(() => {
+            mockStore(sandbox, {
+                app: {
```
alavers: Lots of repeated state here. Tests might be more terse if you returned this state via a `getDefaultState()` helper, and then had each test simply modify the props you want for each scenario.

Lots of code reuse opportunities in general

test/specs/components/action.spec.jsx
lemieux: in fact it doesn't really matter, I do `this.props.currency.toUpperCase()` before passing it

test/specs/components/action.spec.jsx
lemieux: true I'll simplify this

test/specs/services/stripe-service.spec.js
```diff
@@ -0,0 +1,71 @@
+import sinon from 'sinon';
+import { getMock } from 'test/mocks/core';
```
alavers: nit, should maybe call this`createMock`

test/specs/services/stripe-service.spec.js
```diff
@@ -0,0 +1,71 @@
...
+import { getMockedStore } from 'test/utils/redux';
+import * as coreService from 'services/core';
+import { createTransaction, getAccount } from 'services/stripe-service';
+
+const AppStore = require('stores/app-store');
+
+const store = AppStore.store;
+
+function mockStore(s, state = {}) {
```
alavers: Maybe this method should be moved into utils?

test/specs/services/stripe-service.spec.js
lemieux: true it's a mock factory

test/specs/services/stripe-service.spec.js
lemieux: yes

test/specs/components/action.spec.jsx
Mario54: For now, what about adding a `restore` method on the mockedStore returned from the `mockStore` function?

e.g.

```javascript
function mockStore(storeBeingMocked, s, state = {}) {
     var originalStore = storeBeingMocked.store;
     var mockedStore = getMockedStore(s, state);
 
     Object.defineProperty(storeBeingMocked, 'store', {
         get: () => {
             return mockedStore;
         }
     });

     mockedStore.restore = function() {
        Object.defineProperty(storeBeingMocked, 'store', {
            get: () =>
                return originalStore;
           }
     });
 
     return mockedStore;
 }
```

Not sure if that would work or if it's a good idea

lemieux: I'll do something along these lines yes

test/specs/components/action.spec.jsx
lemieux: fixed

test/specs/components/action.spec.jsx
lemieux: fixed

test/specs/services/stripe-service.spec.js
lemieux: fixed

test/specs/services/stripe-service.spec.js
lemieux: done

test/specs/components/action.spec.jsx
lemieux: done

lemieux:
@alavers nit fixed

Mario54:
:+1:

alavers:
:+1:

PR #165, opened 2016-01-20
Cleanup unused animations classes

lemieux:
Seems like these items weren't cleaned up in the rewrite. I also made a change to the default value of `widgetOpened` in the reducer (null instead of undefined).

@mspensieri @dannytranlx

### DIFF
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -25,14 +25,12 @@ import { waitForPage } from 'utils/dom';
 function renderWidget() {
     const el = document.createElement('div');
     el.setAttribute('id', 'sk-holder');
-    el.className = 'sk-noanimation';
 
     const Root = (process.env.NODE_ENV === 'production' ? require('./root-prod') : require('./root-dev')).Root;
     render(<Root store={ store } />, el);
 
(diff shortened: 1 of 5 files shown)


src/js/smooch.jsx
```diff
@@ -25,14 +25,12 @@ import { waitForPage } from 'utils/dom';
 function renderWidget() {
     const el = document.createElement('div');
     el.setAttribute('id', 'sk-holder');
-    el.className = 'sk-noanimation';
```
mspensieri: What was this class originally used for?

mspensieri:
:+1:

src/js/smooch.jsx
lemieux: Before we were applying `sk-close` on the initial render to put it at the bottom, but `sk-close` has an animation.. that was causing the same glitch that was reported. Adding `sk-animation` until it was fully rendered kinda worked around that problem. Now, the initial state of the widget is neither opened or closed, so we don't apply `sk-close` or `sk-appear` and the initial state of the widget is to be at the bottom. So this class is no longer needed.

dannytranlx:
:+1:

PR #166, opened 2016-01-20
Fix npm distribution

### LINKED ISSUE #163: Including smooch via NPM fails due to missing /dist directory
Hi Marc-Antoine Lemieux,

I'm trying to include smooch using browserify. I'm getting the following when executing task:

```
Running "browserify:vendor" (browserify) task
>> Error: Cannot find module 'smooch' from '/vagrant/sails/portal/assets/js'
Warning: Error running grunt-browserify. Use --force to continue.
```

I've taken latest from NPM (2.1.1), packages.config defines main as `dist/smooch.js` which doesn't exist.

lemieux:
This should fix #163 . The problem was that npm is using the .gitignore file if no .npmignore file is provided and was ignoring the dist folder.

I tested it against one of my projects and it worked.

@alavers @mspensieri @dannytranlx


.npmignore
```diff
@@ -0,0 +1,28 @@
+dist/stats.json
```
mspensieri: Would it be easier to just ignore everything, and add reverse rules for the stuff you want to keep?

.npmignore
```diff
@@ -0,0 +1,28 @@
...
+webpack/
+Gruntfile.js
+karma.conf.js
+.babelrc
+.csscomb.json
+.esformatter
+.eslintignore
+.eslintrc
+.jsfmtrc
+bower.json
```
dannytranlx: Add `.jshintrc`?

dannytranlx: And `smooch-js.sublime-project`

.npmignore
dannytranlx: I think @jugarrit tried something like that before without success :confused:

alavers:
:+1:

.npmignore
lemieux: also, this doesn't have a big impact in the end, it's just to be nice and put only what is needed in the npm package... if we forget to ignore stuff later, it won't break anything as long as `dist/smooch.js` is still there.

.npmignore
lemieux: sublime project is covered by `*.sublime-project`, I'll add `jshinrc`

lemieux: fixed

dannytranlx:
:+1:

mspensieri:
:+1:

PR #170, opened 2016-01-25
Return user from init and small event fix

lemieux:
@mspensieri 

as discussed, this returns the app user when you either call `init` or `login`, (and `logout` for that matter).

Also fixed a bug in the events for messages.

### DIFF
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -189,9 +189,12 @@ export class Smooch extends Observable {
             if (!this._el) {
                 this._el = renderWidget();
             }
-            this.trigger('ready');
 
-            return store.getState().user;
+            let user = store.getState().user;
+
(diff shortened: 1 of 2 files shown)


src/js/smooch.jsx
```diff
@@ -190,6 +190,8 @@ export class Smooch extends Observable {
                 this._el = renderWidget();
             }
             this.trigger('ready');
+
+            return store.getState().user;
```
mspensieri: Should we also pass this to the `ready` callback? Is it possible? `this.trigger('ready', store.getState().user)`

mspensieri:
:+1:

src/js/smooch.jsx
lemieux: yes we should

jugarrit:
:+1:

PR #173, opened 2016-01-26
Add default props to message component

### LINKED ISSUE #172: Old messages without actions are crashing the widget
Might be a problem for users with old conversations that had no actions in them.

lemieux:
This would take care of providing an empty array of actions for the old messages.

This would fix #172.

@mspensieri @Mario54


mspensieri:
:+1:

PR #175, opened 2016-02-01
Abort init sequence if running on PhantomJS

mspensieri:
@lemieux

### DIFF
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -96,7 +96,7 @@ export class Smooch extends Observable {
             renderLink();
             this.trigger('ready');
             return Promise.resolve();
-        } else if (/PhantomJS/.test(navigator.userAgent) && !window.SmoochTest) {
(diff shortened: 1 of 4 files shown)


src/js/smooch.jsx
```diff
@@ -96,6 +96,8 @@ export class Smooch extends Observable {
             renderLink();
             this.trigger('ready');
             return Promise.resolve();
+        }else if (/PhantomJS/.test(navigator.userAgent) && !window.SmoochTest) {
```
lemieux: we could use instead `process.env.ENV !== 'test' and inject that value with webpack when we are doing tests.

Also, I was thinking about just adding the PhantomJS string to the regex for the link and that's it.

But yeah, we can just render nothing too.

(and formatting...)

jmtremblay:
:+1:   Side note: I am curious what `if (/lebo|awle|pide|obo|rawli/i.test(navigator.userAgent))` is about. What are those user agents?

alavers:
:+1:

lemieux:
:+1:

PR #176, opened 2016-02-04
B/mobile fixes

alavers:
@lemieux @dannytranlx @Mario54 

Some style fixes that make the UI more usable on mobile

* Fix the text input being half cut off
* The first chat messages grow upwards from the bottom rather than starting at the top (helps with iOS, where the keyboard pushes the whole UI upward)

### DIFF
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -31,11 +31,8 @@ export class ConversationComponent extends Component {
         let messagesHeight = messages.clientHeight;
         let heightRemaining = conversationHeight - (introHeight + messagesHeight + logoHeight);
 
-        if (heightRemaining > logoHeight) {
-            this.refs.logo.className = 'sk-logo anchor-bottom';
-        } else {
-            this.refs.logo.className = 'sk-logo';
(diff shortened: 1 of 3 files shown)


lemieux:
lgtm :+1:

alavers:
what it actually looks like FYI:

demo

src/js/components/conversation.jsx
```diff
@@ -31,11 +31,8 @@ export class ConversationComponent extends Component {
         let messagesHeight = messages.clientHeight;
         let heightRemaining = conversationHeight - (introHeight + messagesHeight + logoHeight);
 
-        if (heightRemaining > logoHeight) {
-            this.refs.logo.className = 'sk-logo anchor-bottom';
```
Mario54: Is `anchor-bottom` unused now?

Mario54:
nice :+1:

src/js/components/conversation.jsx
alavers: Good call, the whole `_positionLogo` helper isn't even necessary at all now.

PR #178, opened 2016-02-08
Use server-side unread count

mspensieri:
@jugarrit @lemieux @alavers 

todo still: 
* bump smooch-core to `0.0.15`
* unit tests? (what should I test?)

### DIFF
--- a/src/js/components/chat-input.jsx
+++ b/src/js/components/chat-input.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { sendMessage, getReadTimestamp, updateReadTimestamp } from 'services/conversation-service';
+import { sendMessage, resetUnreadCount } from 'services/conversation-service';
+import { store } from 'stores/app-store';
 
 
 export class ChatInputComponent extends Component {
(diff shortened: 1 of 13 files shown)


lemieux:
I'd maybe just test that smooch core gets called correctly whenever the reset count happens.

src/js/reducers/conversation-reducer.js
```diff
@@ -1,4 +1,4 @@
-import { ADD_MESSAGE, MESSAGES_CLEARED, SET_CONVERSATION, RESET_CONVERSATION } from 'actions/conversation-actions';
+import * as ConversationActions from 'actions/conversation-actions';
 import { RESET } from 'actions/common-actions';
 
 const INITIAL_STATE = {
```
lemieux: you should add the default unread count here

src/js/components/chat-input.jsx
```diff
@@ -35,8 +36,8 @@ export class ChatInputComponent extends Component {
     }
 
     onFocus(e) {
-        if (getReadTimestamp() > 0) {
-            updateReadTimestamp(0);
+        if (store.getState().conversation.unreadCount > 0) {
```
mspensieri: Am I correct in assuming this should be passed in through props?

Mario54: Yes, you can add the `conversation` state in the `connect` call at the end of the file I think

mspensieri: Hmm this makes it harder to unit test actually

lemieux: If you need the component to rerender when that value changes, it needs to go in  the `connect` part at the bottom, if not, it doesn't matter.

lemieux:
:+1:

PR #181, opened 2016-02-17
expose getConversation

spasiu:
@mspensieri 
@lemieux

### DIFF
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -232,4 +232,8 @@ export class Smooch {
     showSettingsNotification() {
         store.dispatch(showSettingsNotification());
     }
+
+    getConversation() {
+        return getConversation();
+    }
 }
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -234,20 +234,17 @@ export class Smooch {
     }
 
     getConversation() {
-        let conversation = store.getState().conversation;
+        const conversation = store.getState().conversation;
+        const convoNotFoundError = () => {
+            throw new Error('conversation not found');
+        };
+
         if (conversation) {
             return Promise.resolve(conversation);
         }
 
-        return getConversation()
-            .then(response => {
-                if (response.conversation) {
-                    // user.conversationStarted = true
-                    // connect to faye
-                    // write docs and tests
-                    return response.conversation;
-                }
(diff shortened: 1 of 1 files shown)


mspensieri:
Hmm it's a bit more involved than this. The `getConversation` function should:

1. Directly return the conversation if already fetched from the server (i.e. not incur an additional HTTP call)
2. If not fetched yet (because `user.conversationStarted === false`), perform the HTTP call, and handle the response to either:

  2a) reject the promise if conversation not found
  2b) set `user.conversationStarted = true` if found, connect to faye, and return the conversation object

spasiu:
@mspensieri so if I follow the logic correctly, we set conversationStarted server side when the the conversation is got, so I should be all good now.

src/js/smooch.jsx
```diff
@@ -232,4 +232,19 @@ export class Smooch {
...
         store.dispatch(showSettingsNotification());
     }
+
+    getConversation() {
+        const conversation = store.getState().conversation;
+        const convoNotFoundError = () => {
+            throw new Error('conversation not found');
+        };
+
+        if (conversation) {
```
lemieux: that condition will always be true, conversation is never undefined in the store.

src/js/smooch.jsx
```diff
@@ -232,4 +232,19 @@ export class Smooch {
...
+            return Promise.resolve(conversation);
+        }
+
+        return connectFaye()
+            .then(res => res || getConversation())
```
lemieux: don't return `res`, return what's in the store... and a copy of it, not the real object to prevent side effects if they start messing around with it outside of the widget

spasiu: it's not returning res, the next line will pick out the conversation

src/js/smooch.jsx
spasiu: @mspensieri why do we need to explicitly get the conversation in that case?

lemieux: there's a conversationStarted flag on the user... I think it would be better to use that

mspensieri: The use case is:

* No conversation started
* Your backend posts the first message through the REST API
* Call `getConversation()` to tell the SDK that you have started a conversation

it needs to fetch because there is no local copy of the conversation yet

PR #182, opened 2016-02-17
remove onFocus event handler and check count in onChange

spasiu:
@mspensieri 
@lemieux

### DIFF
--- a/src/js/components/chat-input.jsx
+++ b/src/js/components/chat-input.jsx
@@ -17,6 +17,9 @@ export class ChatInputComponent extends Component {
     }
 
     onChange(e) {
+        if (store.getState().conversation.unreadCount > 0) {
+            resetUnreadCount();
+        }
         this.setState({
             text: e.target.value
         });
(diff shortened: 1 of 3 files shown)


src/js/components/chat-input.jsx
```diff
@@ -35,12 +38,6 @@ export class ChatInputComponent extends Component {
         this.refs.input.focus();
     }
 
-    onFocus(e) {
```
lemieux: I don't think we should remove it, both should reset the unread count IMO

src/js/components/chat-input.jsx
```diff
@@ -17,6 +17,9 @@ export class ChatInputComponent extends Component {
     }
 
     onChange(e) {
+        if (store.getState().conversation.unreadCount > 0) {
```
lemieux: write tests for that! :)

lemieux:
:+1:

src/js/components/chat-input.jsx
```diff
@@ -17,11 +17,16 @@ export class ChatInputComponent extends Component {
     }
 
     onChange(e) {
+        checkAndResetUnreadCount();
```
mspensieri: Should we throttle this maybe? `store.getState().conversation.unreadCount` is only reset to `0` after the http call completes - if someone types a lot of characters we're gonna spawn a whole ton of POST requests

lemieux: true, it needs to be throttled, unless we optimisticly put to to 0 before the call and when it comes back it will get updated with the real value?

PR #185, opened 2016-02-23
Remove href on send button

lemieux:
@jpjoyal @alavers @dannytranlx @mspensieri 

We had a report that `preventDefault` was not doing its job in an Angular app and the URL change to `#`.

### DIFF
--- a/src/js/components/chat-input.jsx
+++ b/src/js/components/chat-input.jsx
@@ -54,7 +54,6 @@ export class ChatInputComponent extends Component {
                                value={ this.state.text }></input>
                     </div>
                     <a ref='button'
-                       href='#'
                        className='send'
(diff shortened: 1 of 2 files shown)


mspensieri:
:+1:

dannytranlx:
While you are in that part of code, the input is bleeding into the Send button too, if we change the label.

image

lemieux:
Fix it with CSS ;).. I think the only way to fix that is by using flexbox and it's broken on IE so no go.

dannytranlx:
Ah okai :cry:

dannytranlx:
:+1:

alavers:
dafuq? angular breaks `preventDefault`?

Anyway :+1:

PR #187, opened 2016-02-23
Embedded mode fixes

lemieux:
Notification container was not full width and the logo was misaligned. I also converted the double image retina thing to a single image with a `srcset`.

@jugarrit @Mario54 @dannytranlx @alavers


mspensieri:
:+1:

PR #188, opened 2016-02-23
Events for open and close states

### LINKED ISSUE #180: Observables for open/close?
In the process of tying Smooch to a different button (floating, looks like Intercom), I had to use nasty hacks to determine if the chat widget was open (to hide the button). Being able to observe open/close events (especially started and completed variants) would've helped immensely.

lemieux:
This adds hooks for opening and closing. I also added a `isOpened` method on the Smooch object to know the state. Also, to accomplish this, I added support to call `off` with no arguments to clear all listeners and the widget properly clean up itself when `destroy` is called.

This fixes #180 .

@mspensieri @dannytranlx @alavers @jugarrit

### DIFF
--- a/src/js/services/app-service.js
+++ b/src/js/services/app-service.js
@@ -23,12 +23,10 @@ export function closeWidget() {
 export function toggleWidget() {
     let {embedded, widgetOpened} = store.getState().appState;
     if (!embedded) {
-        store.dispatch(AppStateActions.toggleWidget());
-        
         if (widgetOpened) {
-            observable.trigger('widget:closed');
+            closeWidget();
         } else {
-            observable.trigger('widget:opened');
+            openWidget();
         }
     }
 }
--- a/src/js/services/app-service.js
+++ b/src/js/services/app-service.js
@@ -1,0 +1,34 @@
+import { store } from 'stores/app-store';
+import * as AppStateActions from 'actions/app-state-actions';
+import { observable } from 'utils/events';
+
+
+export function openWidget() {
+    let {embedded} = store.getState().appState;
+    if (!embedded) {
+        store.dispatch(AppStateActions.openWidget());
+        observable.trigger('widget:opened');
+    }
+}
+
+export function closeWidget() {
+    let {embedded} = store.getState().appState;
+    if (!embedded) {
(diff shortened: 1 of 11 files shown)


src/js/services/app-service.js
```diff
@@ -0,0 +1,34 @@
...
+        store.dispatch(AppStateActions.closeWidget());
+        observable.trigger('widget:closed');
+    }
+}
+
+
+export function toggleWidget() {
+    let {embedded, widgetOpened} = store.getState().appState;
+    if (!embedded) {
+        store.dispatch(AppStateActions.toggleWidget());
```
mspensieri: Why not make use of the `openWidget` and `closeWidget` functions?

lemieux: because those are actions creator and shouldn't have side effects

src/js/services/app-service.js
```diff
@@ -0,0 +1,34 @@
+import { store } from 'stores/app-store';
```
mspensieri: What's the purpose of a service? Doesn't the `conversation-service` module handle interfacing with `smooch-core`? This one seems to just hold helper methods

src/js/services/app-service.js
mspensieri: They have just as many side effects as this function does, no? Those functions are dispatching actions to the store, and triggering events on the `observable` like this one is doing

```
export function toggleWidget() {
    let {embedded, widgetOpened} = store.getState().appState;
    if (!embedded) {
        if (widgetOpened) {
            closeWidget();
        } else {
            openWidget();
        }
    }
}
```

src/js/services/app-service.js
lemieux: Initially services were meant to hold real reusable business/data logic. In our webapp we decide to merge them with actions (I should do the change here at some point too). IMO it was the best place to put reusable logic even it doesn't call the server.

src/js/services/app-service.js
lemieux: ohhh I thought you were talking about the action creators... my bad.. you are right, we call those instead.

mspensieri:
:+1: otherwise

spasiu:
:+1:

PR #193, opened 2016-03-02
Width 100% on mobile

lemieux:
This will fix a regression that appeared with the embedded mode. It should go out as soon as possible.

The inner width was stuck at the fixed one instead of taking full available width.

@Mario54  @dannytranlx @mspensieri


Mario54:
:+1:

dannytranlx:
:+1:

mspensieri:
:+1:

PR #194, opened 2016-03-02
Release 2.1.10

lemieux:
# Changes
- Fixes a regression where the widget was not taking full widget on mobile and embedded mode.


mspensieri:
>  the widget was not taking full widget

lolol

PR #196, opened 2016-03-10
Apply new linting rules

lemieux:
I basically just copied over the new linting rules we are using and went through the whole project to fix the errors.

@Mario54 @alavers @mspensieri @dannytranlx


test/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+    "env": {
+        "mocha": true
+    },
+    "globals": {
+        "expect": true
```
Mario54: Do you need to declare expect as global this if you have the mocha env already?

Mario54:
:+1:

Issue #42, opened 2015-07-14
Track URL changes and add updated URL info in ticket.


williamsjj:
If the URL has changed between the time the user last sent a message and the current user message, it would be helpful if SupportKit updated the ticket metadata by adding another metadata event with the new URL.

lemieux:
Closing this since it was merged and deployed today.

PR #200, opened 2016-03-16
Update 'currentUrl' and 'currentTitle' for the device on hash or history change

jugarrit:
Depends on https://github.com/smooch/smooch-core-js/pull/27

@mspensieri 
@dannytranlx 
@jpjoyal 
@alavers 
@lemieux

### DIFF
--- a/src/js/services/user-service.js
+++ b/src/js/services/user-service.js
@@ -8,6 +8,8 @@ let waitForSave = false;
 const waitDelay = 5000; // ms
 let pendingUserProps = {};
 let previousValue = Promise.resolve();
+let deviceUpdateThrottle;
+let deviceUpdatePending = false;
 
 export const EDITABLE_PROPERTIES = [
     'givenName',
@@ -53,15 +55,6 @@ export function update(props) {
     return previousValue;
 }
 
-export function updateNowViewing(deviceId) {
-    return core().appUsers.updateDevice(store.getState().user._id, deviceId, {
-        info: {
-            currentUrl: document.location.href,
-            currentTitle: document.title
-        }
-    });
-}
-
 export function trackEvent(eventName, userProps) {
     const user = store.getState().user;
     return core().appUsers.trackEvent(user._id, eventName, userProps).then((response) => {
@@ -73,4 +66,31 @@ export function trackEvent(eventName, userProps) {
 
         return response;
     });
+}
+
+export function updateNowViewing(deviceId) {
+    if (!deviceUpdateThrottle) {
+        deviceUpdateThrottle = setTimeout(() => {
+            deviceUpdateThrottle = null;
+
+            if (deviceUpdatePending) {
+                updateNowViewing(deviceId);
+                deviceUpdatePending = false;
+            }
+        }, waitDelay);
+
+        return immediateUpdateDevice(deviceId, {
+            info: {
+                currentUrl: document.location.href,
+                currentTitle: document.title
+            }
+        });
+    } else {
+        deviceUpdatePending = true;
+        return Promise.resolve();
+    }
+}
+
+function immediateUpdateDevice(deviceId, device) {
+    return core().appUsers.updateDevice(store.getState().user._id, deviceId, device);
 }
--- a/src/js/services/user-service.js
+++ b/src/js/services/user-service.js
@@ -53,8 +53,13 @@ export function update(props) {
     return previousValue;
 }
 
-export function updateDevice(deviceId, props) {
-    return core().appUsers.updateDevice(store.getState().user._id, deviceId, props);
+export function updateNowViewing(deviceId) {
+    return core().appUsers.updateDevice(store.getState().user._id, deviceId, {
+        info: {
+            currentUrl: document.location.href,
+            currentTitle: document.title
+        }
+    });
 }
 
 export function trackEvent(eventName, userProps) {
--- a/src/js/services/user-service.js
+++ b/src/js/services/user-service.js
@@ -53,6 +53,10 @@ export function update(props) {
     return previousValue;
 }
 
+export function updateDevice(deviceId, props) {
+    return core().appUsers.updateDevice(store.getState().user._id, deviceId, props);
+}
+
 export function trackEvent(eventName, userProps) {
     const user = store.getState().user;
     return core().appUsers.trackEvent(user._id, eventName, userProps).then((response) => {
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -21,7 +21,7 @@ import { getConversation, sendMessage, connectFaye, disconnectFaye, handleConver
 
 import { observable } from 'utils/events';
 import { storage } from 'utils/storage';
-import { waitForPage, monitorUrlChanges, stopMonitoring as stopMonitoringUrlChanges } from 'utils/dom';
+import { waitForPage, monitorUrlChanges, stopMonitoringUrlChanges } from 'utils/dom';
 
 import { Root } from './root';
 
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -16,12 +16,12 @@ import { reset } from 'actions/common-actions';
 import { openWidget, closeWidget } from 'services/app-service';
 import { login } from 'services/auth-service';
 import { getAccount } from 'services/stripe-service';
-import { EDITABLE_PROPERTIES, trackEvent, update as updateUser, updateDevice, immediateUpdate as immediateUpdateUser } from 'services/user-service';
+import { EDITABLE_PROPERTIES, trackEvent, update as updateUser, updateNowViewing, immediateUpdate as immediateUpdateUser } from 'services/user-service';
 import { getConversation, sendMessage, connectFaye, disconnectFaye, handleConversationUpdated } from 'services/conversation-service';
 
 import { observable } from 'utils/events';
 import { storage } from 'utils/storage';
-import { waitForPage } from 'utils/dom';
+import { waitForPage, monitorUrlChanges, stopMonitoring as stopMonitoringUrlChanges } from 'utils/dom';
 
 import { Root } from './root';
 
@@ -159,8 +159,11 @@ export class Smooch {
                 }
             }
         }).then((loginResponse) => {
-            this.monitorUrlChanges();
             store.dispatch(userActions.setUser(loginResponse.appUser));
+
+            monitorUrlChanges(() => {
+                updateNowViewing(getDeviceId());
+            });
 
             if (loginResponse.publicKeys) {
                 store.dispatch(setPublicKeys(loginResponse.publicKeys));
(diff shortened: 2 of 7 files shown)


src/js/smooch.jsx
```diff
@@ -58,7 +58,7 @@ function renderLink() {
 function getDeviceId() {
     const SK_STORAGE = 'sk_deviceid';
     const deviceId = storage.getItem(SK_STORAGE) ||
-        uuid.v4().replace(/-/g, '');
+    uuid.v4().replace(/-/g, '');
```
lemieux: what? indentation is wrong?

src/js/smooch.jsx
```diff
@@ -196,6 +195,40 @@ export class Smooch {
         });
     }
 
+    monitorUrlChanges() {
```
lemieux: this should go in a utils somewhere. It shouldn't be on the public widget API.

src/js/smooch.jsx
```diff
@@ -196,6 +195,40 @@ export class Smooch {
...
+
+        window.history.replaceState = function(state, title, url) {
+            if (url) {
+                this.updateNowViewing();
+            }
+            replaceState && replaceState.apply(window.history, arguments);
+        }.bind(this);
+    }
+
+    updateNowViewing() {
```
lemieux: this should probably be in a service instead.

lemieux:
Also, tests are missing and obviously the smooch-core version bump

src/js/smooch.jsx
alavers: esformatter is removing my indents for lines like this also. hmmm..

lemieux:
This PR will take care of the widget portion of #42

src/js/smooch.jsx
```diff
@@ -196,6 +195,40 @@ export class Smooch {
...
+        return updateDevice(getDeviceId(), {
```
mspensieri: Should we be throttling this like we do for user prop updates?

src/js/smooch.jsx
```diff
@@ -196,6 +195,40 @@ export class Smooch {
         });
     }
 
+    monitorUrlChanges() {
+        const onhashchange = window.onhashchange;
+        const pushState = window.history.pushState;
+        const replaceState = window.history.replaceState;
+
+        window.onhashchange = function() {
```
mspensieri: I think you can use es6 arrow functions, I believe we use babel in this project, @lemieux ?

lemieux: yeah you can and should :)

lemieux: actually these should probably use this format : `window.addEventListener("hashchange", fn, false);`

That's the way it's done for other global events in this project and I'd prefer we stick we the same convention.

src/js/smooch.jsx
lemieux: No I don't think we should because I don't think this should be public facing at all.

src/js/smooch.jsx
jugarrit: Can you do that for pushState?

jugarrit: Also when I used arrow functions it didn't work...

Anything look wrong here?

```
        window.onhashchange = () => {
            this.updateNowViewing();
            onhashchange && onhashchange.apply(window, arguments);
        };

        window.history.pushState = (state, title, url) => {
            if (url) {
                this.updateNowViewing();
            }
            pushState && pushState.apply(window.history, arguments);
        };

        window.history.replaceState = (state, title, url) => {
            if (url) {
                this.updateNowViewing();
            }
            replaceState && replaceState.apply(window.history, arguments);
        };
```

jugarrit: Results in this error: 
uncaught TypeError: Failed to execute 'pushState' on 'History': 2 arguments required, but only 0 present

So the arguments are empty or something...

lemieux: you can try 
```
(...args) => {
            this.updateNowViewing();
            onhashchange && onhashchange.apply(window, args);
        };
```

jugarrit: Sure, I could try that, but why wouldn't the original work?

src/js/smooch.jsx
jugarrit: But it's in the public API, so it is public facing.

src/js/smooch.jsx
jugarrit: The `updateDevice` portion is in a service, why does this need to be?

src/js/smooch.jsx
lemieux: usually passing `arguments` works in some case but the real way of doing it would be to coerce it to an array (arguments is array-ish). Try wrapping it in `Array.from`to see if it works

src/js/smooch.jsx
lemieux: unless we want to expose it, but there isn't much to be done by a user here.

src/js/smooch.jsx
lemieux: as discussed, you can merge the logic of getting the device id in the service and just hide the implementation in the service. The service holds the business logic.

src/js/smooch.jsx
lemieux: > Can you do it for pushState? 

Actually you can listen to the popstate event on `window` (https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_popstate_event) instead of listening to replaceState and pushState

then you would window.addEventListener for both hashchange and popstate.

lemieux: Also it would prevent the widget from crashing if `window.history` doesn't exist for some reason.

src/js/smooch.jsx
mspensieri: But the hash can change very frequently, no? If someone scrolls down on docs.smooch.io, the hash changes 20 times / sec

src/js/smooch.jsx
jugarrit: But that won't work based on the SO article you posted:

http://stackoverflow.com/questions/4570093/how-to-get-notified-about-changes-of-the-history-via-history-pushstate

5.5.9.1 Event definitions

"The popstate event is fired in certain cases when navigating to a session history entry.
According to this, there is no reason for popstate to be fired when you use pushState. But an event such as pushstate would come in handy. Because history is a host object, you should be careful with it, but Firefox seems to be nice in this case."

lemieux: okay, true! I can't believe people need to rely on monkey patching to do these kind of things O.o

lemieux: then put a check around this for the existence of `window.history` (IE8-IE9 don't support that, but we don't support them either... but at least we can prevent it from crashing)

src/js/smooch.jsx
dannytranlx: Do we have a line-break rule? We could just keep it on one line for now!

alavers:
Rest looks :+1:

src/js/smooch.jsx
alavers: Looks like a sublime-esformatter problem.. :(

Works in command line esformatter
```
▶ <test.js
'use strict';

let blap = 42 ||
100;

function foo() {}

let bar = foo() ||
96;

~/git/agent-console  b/cox-email-marked ✗                                                                                                            11m ◒
▶ esformatter test.js
'use strict';

let blap = 42 ||
    100;

function foo() {
}

let bar = foo() ||
    96;
```

alavers: everyone switch to vim :trollface:

jugarrit: I'm using atom!

dannytranlx: Atom does the same :confused: 
atom-esformatter

alavers: ok wtf I tinkered with my esformatter version and now vim has busted behavior too.

alavers: OK mystery solved. esformatter@0.8.1 does not exhibit this fucked behavior.

alavers: I think keep using latest esformatter, we'll just tolerate the fucky indent for now. or save without esformatter if it really bothers you

src/js/smooch.jsx
jugarrit: Will address this.

src/js/smooch.jsx
```diff
@@ -16,12 +16,12 @@ import { reset } from 'actions/common-actions';
...
 import { login } from 'services/auth-service';
 import { getAccount } from 'services/stripe-service';
-import { EDITABLE_PROPERTIES, trackEvent, update as updateUser, immediateUpdate as immediateUpdateUser } from 'services/user-service';
+import { EDITABLE_PROPERTIES, trackEvent, update as updateUser, updateNowViewing, immediateUpdate as immediateUpdateUser } from 'services/user-service';
 import { getConversation, sendMessage, connectFaye, disconnectFaye, handleConversationUpdated } from 'services/conversation-service';
 
 import { observable } from 'utils/events';
 import { storage } from 'utils/storage';
-import { waitForPage } from 'utils/dom';
+import { waitForPage, monitorUrlChanges, stopMonitoring as stopMonitoringUrlChanges } from 'utils/dom';
```
Mario54: Why not call `stopMonitoringUrlChanges` in the export instead of changing the name when you import it?

Mario54:
Rest is :+1:

src/js/smooch.jsx
jugarrit: k

src/js/smooch.jsx
alavers: Fix incoming :D 

https://github.com/millermedeiros/esformatter/pull/422

alavers: https://github.com/millermedeiros/esformatter/pull/422

src/js/services/user-service.js
```diff
@@ -65,3 +67,30 @@ export function trackEvent(eventName, userProps) {
         return response;
     });
 }
+
+export function updateNowViewing(deviceId) {
+    if (!deviceUpdateThrottle) {
```
lemieux: did you reuse the same throttling logic from update user?

mspensieri:
:+1:

src/js/services/user-service.js
jugarrit: No, I didn't. There was a portion I didn't quite understand, and it was mixed in with 'isDirty', which I'm already checking in a different way.

jugarrit: Another reason I didn't use it... There's a bug

jugarrit: It doesn't call immediateUpdate at the tail end of the throttle if there is an update queued.

lemieux: I think it was by design, but maybe not. User is force-saved before sending a message anyway. But yeah might be a good idea to do the tail call.. Can you add it? pls

lemieux: nevermind, just saw the merge notif. I'll do it in my branch.

PR #201, opened 2016-03-18
Refresh the conversation on web socket connect

jugarrit:
@mspensieri 
@lemieux 
@jpjoyal 
@Mario54 
@dannytranlx

### DIFF
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -164,14 +164,6 @@ export class Smooch {
                 }
             });
         }).then((loginResponse) => {
-            observable.on('socket:connected', () => {
-                const user = store.getState().user;
-
-                if (user.conversationStarted) {
-                    getConversation();
-                }
-            });
-
             store.dispatch(setUser(loginResponse.appUser));
 
             if (loginResponse.publicKeys) {
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -129,8 +129,6 @@ export class Smooch {
         store.dispatch(resetUser());
         store.dispatch(resetConversation());
 
-        observable.off('socket:connected');
-
         disconnectFaye();
 
         attributes = pick(attributes, EDITABLE_PROPERTIES);
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -128,6 +128,9 @@ export class Smooch {
         store.dispatch(resetAuth());
         store.dispatch(resetUser());
         store.dispatch(resetConversation());
+
+        observable.off('socket:connected');
+
         disconnectFaye();
 
         attributes = pick(attributes, EDITABLE_PROPERTIES);
@@ -161,6 +164,14 @@ export class Smooch {
                 }
             });
         }).then((loginResponse) => {
+            observable.on('socket:connected', () => {
+                const user = store.getState().user;
+
+                if (user.conversationStarted) {
+                    getConversation();
+                }
+            });
+
             store.dispatch(setUser(loginResponse.appUser));
 
             if (loginResponse.publicKeys) {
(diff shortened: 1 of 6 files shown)


src/js/smooch.jsx
```diff
@@ -161,6 +164,14 @@ export class Smooch {
                 }
             });
         }).then((loginResponse) => {
+            observable.on('socket:connected', () => {
```
lemieux: The login promise should resolve once this is triggered IMO

lemieux: Also, is it useful to publish that on the public event hub.. And why use events instead of calling the necessary functions directly... I don't want to fall back in the events trap Backbone and Marionette had us do.

jugarrit: I'm really never sure what's preferable honestly.
I feel like sometimes I try to call the method directly and am told to use events, and other times the opposite is true.
I don't mind either way :)

jugarrit: I don't know if I agree that is should resolve when this is triggered, sometimes it'll get triggered and it's unrelated to login

jugarrit: Also what's the best way to notify Smooch from Faye if not through this hub? Eventually we want to listen to socket:closed to show an error message in the UI while Faye is disconnected.

lemieux: I'd say use the least events possible with React and Redux.. The app will react to a change in the store if needed anyway.. Thoughts @alavers?

jugarrit: Would we simply dispatch a conversation action from `faye.js`?

lemieux: I think so yeah

alavers: Yeah I would prefer dispatching all of these events via faye. No half measures.

image

jugarrit: I'm not trying to cut corners, I just genuinely can't ever think of the correct path to take when I'm coding in redux. I find it extremely confusing.

jugarrit: But also wouldn't calling it directly be cutting corners?

alavers: I'd be OK with faye.js directly calling conversation-service.getConversation

faye.js is already playing the role of gluing faye events to redux dispatching, so it makes sense.

Not about cutting corners, it's about embracing redux :D

jugarrit: Then I shall embrace the reduction with open arms.

alavers:
:+!

test/specs/smooch.spec.js
```diff
@@ -113,6 +113,12 @@ describe('Smooch', () => {
             disconnectFayeStub = sandbox.stub(conversationService, 'disconnectFaye');
         });
 
+        afterEach(() => {
+            loginStub.restore();
```
Mario54: Aren't these covered by `sandbox.restore()` on line 60?

test/specs/faye/faye.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    describe('when connected', () => {
+        beforeEach(() => utilsFaye.initFaye());
+
+        describe('when conversation is started', () => {
```
Mario54: Weren't you going to add a test to check that `getConversationStub` is not called if `conversationStarted` is false?

test/specs/smooch.spec.js
jugarrit: Yeah I added these when I thought line 60 wasn't working

test/specs/faye/faye.spec.js
jugarrit: Yep! Just didn't finish

Mario54:
:+1: after you add the test :)

src/js/smooch.jsx
```diff
@@ -128,6 +128,9 @@ export class Smooch {
         store.dispatch(resetAuth());
         store.dispatch(resetUser());
         store.dispatch(resetConversation());
+
+        observable.off('socket:connected');
```
lemieux: is this still valid?

lemieux:
:+1: after tests

PR #189, opened 2016-02-24
expose getConversation

spasiu:
- [x] working on unit tests

PR #202, opened 2016-03-20
Do not overwrite user properties on getConversation

### LINKED ISSUE #199: getConversation method reset the appUser and return 0 unread message all the time
Hy,
I have integrates smooch on our webapp, it's really cool! but i'm trying to open by default the tchat after the first load of the application if we have unread messages.

```
Smooch.init({
  appToken: 'xyz'
});

Smooch.on('ready', function () {
  Smooch.getConversation().then(function (conversation) {
    if (conversation.unreadCount > 0) {
      Smooch.open();
    }
  });
});
```

it's look like really simply, but if we run this we get errors when we sending messages.

capture d ecran 2016-03-16 a 19 47 41

so... ok I don't know why a getConversation reset the userApp settings but I tried to fix this by add

```
Smooch.on('ready', function () {
  
  [...]

  Smooch.updateUser({ [...] });
});
```

But at this time `unreadCount` was always set at `0`

Thanks in advance for your response 🦄

Mario54:
@spasiu @lemieux @alavers 

Fixes #199, a bug introduced in #189. 

When `getConversation` succeeded, we were overwriting user properties, which means it became empty apart from having a `conversationStarted` property of true.

### DIFF
--- a/src/js/actions/user-actions.js
+++ b/src/js/actions/user-actions.js
@@ -1,10 +1,18 @@
 export const SET_USER = 'SET_USER';
+export const UPDATE_USER = 'UPDATE_USER';
 export const RESET_USER = 'RESET_USER';
 
 export function setUser(props) {
     return {
         type: SET_USER,
         user: props
+    };
+}
+
+export function updateUser(properties) {
+    return {
+        type: UPDATE_USER,
+        properties
     };
 }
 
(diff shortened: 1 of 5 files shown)


lemieux:
:+1:

dannytranlx:
Nice, thanks for the fix and tests :grin: 
:+1:

mspensieri:
:+1:

alavers:
:+1:

PR #203, opened 2016-03-21
F/use colors

mspensieri:
@Mario54 @lemieux @alavers @jugarrit 

Use colors returned in the `init` response. Will wait to merge until color picker is deployed


src/js/components/conversation.jsx
```diff
@@ -53,7 +53,10 @@ export class ConversationComponent extends Component {
     }
 
     render() {
-        const messages = this.props.conversation.messages.map((message) => <MessageComponent key={ message._id } {...message} />);
+        const messages = this.props.conversation.messages.map((message) => <MessageComponent key={ message._id }
```
alavers: Nit: Ow my code editor. Can we put jsx here on a new line? Eg:

```
        const messages = this.props.conversation.messages.map((message) => (
            <MessageComponent key={ message._id }
                              accentColor={ this.props.settings.accentColor }
                              linkColor={ this.props.settings.linkColor }
                              {...message} />
        ));
```

src/js/components/conversation.jsx
```diff
@@ -67,11 +70,15 @@ export class ConversationComponent extends Component {
...
+                    <div ref='messages'
+                         className='sk-messages'>
                         { messages }
                     </div>
-                    <div className='sk-logo' ref='logo' style={logoStyle}>
-                        <a href='https://smooch.io/?utm_source=widget' target='_blank'><span>In-App Messaging by</span> <img className='sk-image'
+                    <div className='sk-logo'
+                         ref='logo'
+                         style={ logoStyle }>
+                        <a href='https://smooch.io/?utm_source=widget'
```
alavers: esformatter gone berzerk?

src/js/components/chat-input.jsx
```diff
@@ -113,7 +120,8 @@ export class ChatInputComponent extends Component {
 
 export const ChatInput = connect((state) => {
     return {
-        ui: state.ui
+        ui: state.ui,
+        settings: state.app.settings.web
```
alavers: Should this tolerate an undefined settings?

src/js/components/conversation.jsx
alavers: And actually you can be even fancier if you wish:

```
        const messages = this.props.conversation.messages.map((message) => (
            <MessageComponent key={ message._id }
                              {...this.props.settings}
                              {...message} />
        ));
```

src/js/components/notification.jsx
```diff
@@ -35,10 +39,15 @@ export class NotificationComponent extends Component {
...
         return (
-            <div key='content' className='sk-notification' onClick={ this.props.actions.hideSettingsNotification }>
+            <div key='content'
+                 className='sk-notification'
+                 onClick={ this.props.actions.hideSettingsNotification }>
                 <p>
-                    <span ref='text' dangerouslySetInnerHTML={ createMarkup(this.props.ui.text.settingsNotificationText) }></span>
-                    <a href='#' className='sk-notification-close' onClick={ this.props.actions.hideSettingsNotification }>&times;</a>
+                    <span ref='text'
+                          dangerouslySetInnerHTML={ createMarkup(this.props.ui.text.settingsNotificationText) }></span>
```
alavers: I'm not sure I like this createMarkup helper. It's only hiding the `{ __html: html }` crap which is supposed to be ugly by design.

lemieux: I don't mind writing the wrapper everytime but `dangerouslySetInnerHTML` is ugly enough

src/js/components/conversation.jsx
lemieux: ^ :+1:

src/js/components/conversation.jsx
lemieux: no it hit the 2 attr limit on one tag before breaking I guess

src/stylesheets/buttons.less
```diff
@@ -27,6 +27,7 @@
   &.focus {
     color: @btn-default-color;
     text-decoration: none;
+    .opacity(.8);
```
lemieux: neat hack

src/js/components/chat-input.jsx
```diff
@@ -86,9 +86,14 @@ export class ChatInputComponent extends Component {
         let sendButton;
 
         const buttonClassNames = ['send'];
+        const buttonStyle = {};
 
         if (this.state.text.trim()) {
             buttonClassNames.push('active');
+
+            if (this.props.settings.accentColor) {
+                buttonStyle.color = `#${this.props.settings.accentColor}`;
```
lemieux: text color is always white?

alavers:
Rest looks good

src/js/components/conversation.jsx
mspensieri: Don't we have a formatter rule to force no-newline?

src/js/components/chat-input.jsx
mspensieri: The action creator makes sure it's never undefined

src/js/components/chat-input.jsx
mspensieri: Not on the Send button

src/js/components/chat-input.jsx
mspensieri: ... but it's failing a ton of unit tests, so I'll fix

src/js/components/chat-input.jsx
lemieux: hmm sorry comment was placed at the wrong spot

src/js/components/action.jsx
```diff
@@ -62,6 +62,11 @@ export class ActionComponent extends Component {
     render() {
         const publicKeys = store.getState().app.publicKeys;
 
+        let style = {};
+        if (this.props.accentColor) {
+            style.backgroundColor = style.borderColor = `#${this.props.accentColor}`;
```
lemieux: how do you define the text color? always white?

mspensieri: Yeah always white

lemieux: Same for the bubble I guess? In that case it might be clashing and unreadable in some case... we might want to detect if it's a dark or light color and choose the text color based on that.

mspensieri: It's the app maker's responsibility to choose colors that don't clash - works the same way in iOS.

lemieux: I guess but how are they supposed to choose that color when using the color settings?

src/js/components/chat-input.jsx
lemieux: and actually, here the button is always blue but the text color changes?

mspensieri: hover

lemieux: wth... `buttonStyle.color` changes the background color?

src/js/components/chat-input.jsx
```diff
@@ -128,7 +135,8 @@ export class ChatInputComponent extends Component {
 
 export const ChatInput = connect((state) => {
     return {
-        ui: state.ui
+        ui: state.ui,
+        settings: (state.app.settings && state.app.settings.web) || {}
```
lemieux: Maybe use a defaultProp in the component instead of using `|| {}`?

src/js/components/chat-input.jsx
mspensieri: haha no, this code is only the send button which uses `color`, the action buttons use `background-color` in `src/js/components/action.jsx`

lemieux: ugh, didn't check where this comment was..... you scared me for a minute :p

lemieux:
:+1: works for me if that's the expected behavior for the text color

lemieux:
still works for me :+1:

PR #207, opened 2016-03-21
Retry resizing input if widget has no width

lemieux:
When using the embed mode, it can happen that the widget has no width on the initial render. This would retry getting the width 10 times in 1 sec and then default to 80% width. The input is resize whenever the page is resized after that.

@alavers @dannytranlx @Mario54 @jugarrit


Mario54:
:+1:

alavers:
:+1:

dannytranlx:
:

PR #209, opened 2016-03-22
Image upload support

lemieux:
Here you go guys, image upload!

I still need to add the documentation.

@alavers @mspensieri @Mario54 @dannytranlx @jugarrit


src/js/services/conversation-service.js
```diff
@@ -39,22 +58,51 @@ export function sendMessage(text) {
...
             return response;
-        }).then(handleFirstUserMessage);
-    };
+        });
+    });
+}
 
-    var promise = immediateUpdate(store.getState().user);
+export function uploadImage(file) {
+    if (isFileTypeSupported(file.type)) {
```
mspensieri: nit: 

```
if (!isFileTypeSupported(file.type)) {
  store.dispatch(showErrorNotification(store.getState().ui.text.invalidFileError));
  return Promise.reject('Invalid file type');
}

 return resizeImage(file).then((dataUrl) => {
// .....
```

mspensieri:
:+1: needs bug bash :)

src/js/services/conversation-service.js
lemieux: hmmm why checking the false case instead of checking the true case? Early exit?

mspensieri: Yeah it's a bit nicer to exit early so you don't have a giant `if` block

lemieux: /shrug I prefer to evaluate true case first but sure I'll change

src/js/components/chat-input.jsx
```diff
@@ -70,7 +78,7 @@ export class ChatInputComponent extends Component {
     }
 
     componentDidMount() {
-        this.resizeInput();
+        setTimeout(() => this.resizeInput());
```
dannytranlx: ?

lemieux: this makes sure to execute  `this.resizeInput` after the current stack, so usually after the HTML was actually flushed to the DOM and the widget has actual height and width (needed to resize the input)

src/js/smooch.jsx
```diff
@@ -103,6 +108,12 @@ export class Smooch {
             store.dispatch(AppStateActions.disableSettings());
         }
 
+        if(props.imageUploadEnabled && isImageUploadSupported()) {
```
dannytranlx: nit: formatting?

README.md
```diff
@@ -71,6 +71,7 @@ Initializes the Smooch widget in the web page using the specified options. It re
 | userId | Yes | - | User's id |
 | properties | Yes | - | An object with all properties you want to set on your user |
 | emailCaptureEnabled | Yes | `false` | Enables prompt for email after the first user's message. You can retrieve that email in Slack using `/sk !profile`. We are aware of this limitation and are working on improving it. |
+| imageUploadEnabled | Yes | `true` | Enables the image upload feature. |
```
Mario54: Should we document that in our docs, or in the README is good enough?

src/js/components/chat-input.jsx
```diff
@@ -7,6 +7,8 @@ import debounce from 'lodash.debounce';
 import { sendMessage, resetUnreadCount } from 'services/conversation-service';
 import { store } from 'stores/app-store';
 
+import { ImageUpload } from 'components/image-upload';
```
Mario54: Should this be a default export?

lemieux: Widget doesn't do default exports.

README.md
lemieux: I was looking through the existing docs and it's not even mentionned for iOS.

Mario54: I don't think it's possible to disable the image upload. @mspensieri ?

Mario54:
:+1:

make-webpack-config.js
```diff
@@ -27,8 +27,7 @@ module.exports = function(options) {
         'jsx': options.hotComponents ? ['react-hot-loader', 'babel-loader'] : 'babel-loader',
         'js': {
             loader: 'babel-loader',
-            include: [path.join(__dirname, 'src/js'), path.join(__dirname, 'test')],
-            exclude: [/node_modules/]
+            include: [path.join(__dirname, 'src/js'), path.join(__dirname, 'test')]
```
alavers: What magic is this? exclude no longer necessary?

lemieux: from what I understand, exclude was never necessary because we use `include`

package.json
```diff
@@ -114,6 +118,7 @@
     "phantomjs-prebuilt": "2.1.4",
     "react-addons-test-utils": "0.14.3",
     "react-hot-loader": "1.3.0",
+    "react-imageloader": "2.1.0",
```
alavers: What determines if a `react-*` module is a dependency or a devDependency? Are all of those `react-*` modules above runtime deps?

src/js/components/chat-input.jsx
alavers: remind me why?

README.md
mspensieri: Nope, not possible. Been requested a couple times though

package.json
lemieux: ehh image loader should definitely be in prod deps... those that are in devDeps are either only used for tests or in the webpack process (like react-hot-loader)

src/js/components/chat-input.jsx
lemieux: Because I built it that way before we even started on the web app. We could go through everything and fix it though, but I don't think it's necessary in this PR.

README.md
lemieux: IMO we should consider the image upload as a de facto feature, but if people ask how to disable it, we have an answer.

src/stylesheets/footer.less
```diff
@@ -9,12 +9,39 @@
...
+    }
+
     .input-container {
         width: 100%;
-        padding: 5px 0 5px @input-horizontal-margin;
-        float: left;
+        padding: 5px 0;
+        display: inline-block;
+
+        &.no-upload {
```
alavers: feels odd to me that we need both this css and the javascript compute button/input widths.

Can this width computation not be done purely in CSS?

lemieux: if we were using `display: flex` it would be done automatically. We need to do it with javascript for the cases where people change the send button text and it becomes larger.

src/js/components/conversation.jsx
```diff
@@ -53,7 +53,7 @@ export class ConversationComponent extends Component {
     }
 
     render() {
-        const messages = this.props.conversation.messages.map((message) => <MessageComponent key={ message._id } {...message} />);
+        const messages = this.props.conversation.messages.map((message, index) => <MessageComponent key={ index } onLoad={this.scrollToBottom} {...message} />);
```
alavers: of we scroll for each message `onLoad` now, do we still need the componentDidMount and componentDidUpdate triggers as well?

make-webpack-config.js
alavers: Ah, so then `exclude` here only applies to the `include` subset

src/js/components/conversation.jsx
lemieux: we do need them for normal messages and for when a conversation doesn't have images in it

src/js/components/chat-input.jsx
alavers: :+1:

make-webpack-config.js
lemieux: or doesn't apply at all, I'm still not sure about it

src/js/components/image-message.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { Component } from 'react';
+import ImageLoader from 'react-imageloader';
+
+
+function Loading() {
```
alavers: `loading.jsx`?

lemieux: sure

PR #210, opened 2016-03-22
Fix rounding issue for input width

lemieux:
As seen with the widget embedded in Elev.io, the input was just 1px to wide and was pushing the send button out. I realize that `offsetWidth` is always returning a rounded integer. In Elev.io case, the actual width (using the client rect) is 49.390625, but offsetWidth returns 49. This is just enough to break the footer rendering.

By using `getBoundingClientRect`, we ensure the real computed width is used and then we ceil it to get the worst case scenario.

@alavers @dannytranlx @jugarrit @mspensieri

### DIFF
--- a/src/js/components/chat-input.jsx
+++ b/src/js/components/chat-input.jsx
@@ -38,9 +38,17 @@ export class ChatInputComponent extends Component {
 
     resizeInput(attempt = 0) {
         const node = findDOMNode(this);
-        if (node.offsetWidth - this.refs.button.offsetWidth > 0) {
+
+        const nodeRect = node.getBoundingClientRect();
+        const buttonRect = this.refs.button.getBoundingClientRect();
+
+        // use floor on widget and ceil on button to get worst rounding scenario
+        const nodeWidth = Math.floor(nodeRect.width);
+        const buttonWidth = Math.ceil(buttonRect.width);
+
+        if (nodeWidth - buttonWidth > 0) {
             this.setState({
-                inputContainerWidth: node.offsetWidth - this.refs.button.offsetWidth
+                inputContainerWidth: nodeWidth - buttonWidth
             });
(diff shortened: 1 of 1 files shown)


src/js/components/chat-input.jsx
```diff
@@ -38,9 +38,17 @@ export class ChatInputComponent extends Component {
 
     resizeInput(attempt = 0) {
         const node = findDOMNode(this);
-        if (node.offsetWidth - this.refs.button.offsetWidth > 0) {
+
+        const nodeRect = node.getBoundingClientRect();
+        const buttonRect = this.refs.button.getBoundingClientRect();
+
+        // use floor on widget and ceil on button to get worst rounding scenario
```
jugarrit: Hein?

dannytranlx: So the `ceil` is to _overshoot_, just to make sure we have enough width?

lemieux: find the minimum width the widget could take when it's a decimal, find the maximum width the button can take when it's a decimal.. that way, widget width - button width is always smaller or equal than the actual width

dannytranlx:
:+1:

src/js/components/chat-input.jsx
jugarrit: I see...
The comment is a bit confusing.
Maybe something like... // floor widget width and ceil button width to ensure button width <= widget width
Why can't they be decimals, though?

lemieux: Because not all browser interpret them the same. and I'll use your comment

lemieux: And I prefer to be on the safe side and not trust the browser to round the value to what it needs to be.

jugarrit:
:+1:

Issue #212, opened 2016-03-22
Dependency conflicts?


mtwilliams:
I'm not using React, or depending on anything that uses React. Verified it's `smooch` by removing it from `package.json` and doing a fresh install of dependencies.

```
$ npm install --save
npm WARN package.json walter@ No license field.
npm ERR! Darwin 14.5.0
npm ERR! argv "/usr/local/Cellar/node/4.2.1/bin/node" "/usr/local/bin/npm" "install" "--save"
npm ERR! node v4.2.1
npm ERR! npm  v2.14.7
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package react@0.14.2 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer react-addons-css-transition-group@0.14.3 wants react@^0.14.3
npm ERR! peerinvalid Peer react-dom@0.14.2 wants react@^0.14.2
npm ERR! peerinvalid Peer react-redux@4.0.5 wants react@^0.14.0
```

lemieux:
Thanks for the report, which version of npm are you using?

lemieux:
Oh sorry it's there right in my face. You're using npm 2.14.7 :) 

I guess I'm using npm 3.0 for too long now! I'll publish a fix that will take care of these dependencies and should work with npm 2.

mtwilliams:
:raised_hands:

lemieux:
@mtwilliams version 2.1.15 was just released and should fix those dependency problems!

PR #213, opened 2016-03-22
Fix invalid peer dependencies on npm 2

lemieux:
This should take care of #212. Dependency conflicts still exists in dev dependencies though, due to us using sinon@2.0.0-pre and sinon-as-promised, but there's nothing we could really do about it. This PR would at least fix the prod dependencies.

I also simplified one test that was failing for the settings component for an unknown reason. I tested it manually and it's still working correctly.

@alavers @mspensieri @dannytranlx

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,4 @@
-    {
+{
   "name": "smooch",
   "analyze": false,
   "version": "2.1.14",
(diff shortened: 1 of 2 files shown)


mspensieri:
:+1:

Mario54:
:+1:

PR #216, opened 2016-03-23
Block touch scroll only when not using embedded

lemieux:
This will allow user to scroll past the conversation when using embedded widget.

@jpjoyal @Mario54 @mspensieri @dannytranlx


Mario54:
:+1:

mspensieri:
:+1:

PR #221, opened 2016-03-25
Bump font awesome version

RHavar:
If an old version of font-awesome is loaded on a page that uses a new version of font-awesome, all the new font-awesome icons get unloaded


lemieux:
Hey @RHavar, sorry about that awfully long response (was on vacation, and wtf is wrong with you team!)

This seems reasonable :) I'll get that published in our next release!

Thanks a lot !

PR #223, opened 2016-04-06
Trigger message event on message coming from getConversation

lemieux:
Events weren't triggered when a new message was coming from getConversation or straight from init. Now appMaker messages and whispers are handled by computing which messages are new based on the unread count.

App user message are still triggered when the user actually send  them and this change doesn't touch that behavior.

@dannytranlx @mspensieri @alavers @Mario54

### DIFF
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -41,6 +41,22 @@ export class Observable {
 
 export const observable = new Observable();
 
+export function observeStore(store, select, onChange) {
+    let currentState;
+
+    function handleChange() {
+        const nextState = select(store.getState());
+        if (nextState !== currentState) {
+            currentState = nextState;
+            onChange(currentState);
+        }
+    }
+
+    const unsubscribe = store.subscribe(handleChange);
+    handleChange();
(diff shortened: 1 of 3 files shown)


src/js/utils/events.js
```diff
@@ -41,6 +41,22 @@ export class Observable {
 
 export const observable = new Observable();
 
+export function observeStore(store, select, onChange) {
+    let currentState;
+
+    function handleChange() {
+        const nextState = select(store.getState());
+        if (nextState !== currentState) {
```
mspensieri: should this trigger only on changes related to the conversation store?

lemieux: this function is generic, the `select` param is a function that specify what is interesting in the store (and if you look at https://github.com/smooch/smooch-js/pull/223/files#diff-b089bf546efa30553b8f38a397c6c748R144, I'm only interested in the conversation key)

mspensieri: Ah, I missed that part

mspensieri:
:+1:

Mario54:
👍

PR #230, opened 2016-04-07
Set conversationStarted to true after sending a message successfully

Mario54:
@lemieux @dannytranlx @mspensieri @alavers 

After a first user message, if there was a whisper or a out-of-office reply added to the conversation it would not show up in the Web SDK because it was dispatched to faye before the widget could connect. And, since `conversationStarted` was false, it did not fetch the conversation when the connection to faye was established, which means we lost messages between the message being sent and connecting to faye.


mspensieri:
:+1:

alavers:
:+1:

lemieux:
:+1:

PR #231, opened 2016-04-08
Check for existing message without id when adding messages

lemieux:
A small bug caused sent messages to appear twice for a fraction of a second when it was received by faye.

This change will ensure the new message won't get added to the conversation.

@mspensieri @Mario54 @dannytranlx

### DIFF
--- a/src/js/reducers/conversation-reducer.js
+++ b/src/js/reducers/conversation-reducer.js
@@ -20,7 +20,12 @@ const sortMessages = (messages) => messages.sort((a, b) => {
 });
 
 const addMessage = (messages, message) => {
-    const existingMessage = messages.find((m) => m._id === message._id);
+    let existingMessage = messages.find((m) => m._id === message._id);
+
+    // let's try to match against recently sent messages instead
+    if (!existingMessage) {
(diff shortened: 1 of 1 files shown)


Mario54:
:+1:

src/js/reducers/conversation-reducer.js
```diff
@@ -20,7 +20,12 @@ const sortMessages = (messages) => messages.sort((a, b) => {
 });
 
 const addMessage = (messages, message) => {
-    const existingMessage = messages.find((m) => m._id === message._id);
+    let existingMessage = messages.find((m) => m._id === message._id);
+
+    // let's try to match against recently sent messages instead
+    if (!existingMessage) {
```
mspensieri: Could check for `role === 'appUser'` here as a small optimization. App maker messages that come in through faye shouldn't suffer from this bug

mspensieri:
:+1:

src/js/reducers/conversation-reducer.js
lemieux: true, I'll add that

PR #233, opened 2016-04-11
Blink title when new messages come in

### LINKED ISSUE #40: Favicon or page title "reply waiting" indicator would be great.
It would be awesome if the favicon or page title could flash when there's a reply waiting for the user, so when they're multitasking they can know there's something ready for them.

lemieux:
This should fix #40 .

The title of the page will alternate between the original title and "{User} said...". This should draw attention to the tab. 

@dannytranlx @Mario54 @mspensieri

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ var skPromise = Smooch.init({
         actionPaymentCompleted: 'Payment Completed',
         messageError: 'An error occured while sending your message. Please try again.',
         invalidFileError: 'Sorry, but only images are supported currently. Please choose a file with a supported extension (jpg, jpeg, png, gif, or bmp).',
-        messageIndicatorTitle: '{name} said...'
+        messageIndicatorTitleSingular: '({count}) New message',
+        messageIndicatorTitlePlural: '({count}) New messages'
     }
 });
 
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Initializes the Smooch widget in the web page using the specified options. It re
 | emailCaptureEnabled | Yes | `false` | Enables prompt for email after the first user's message. You can retrieve that email in Slack using `/sk !profile`. We are aware of this limitation and are working on improving it. |
 | imageUploadEnabled | Yes | `true` | Enables the image upload feature. |
 | embedded | Yes | False | Tells the widget it will be embedded. (see Embedded section below) |
-| customText | Yes | See the example below | Strings used in the widget UI. You can use these to either customize the text or translate it. |
+| customText | Yes | See the example below | Strings used in the widget UI. You can use these to either customize the text or translate it. If something is between `{}`, it's a variable and needs to stay in your customized text if you want to use it. |
 
 ```javascript
 var skPromise = Smooch.init({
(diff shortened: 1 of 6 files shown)


src/js/components/message-indicator.jsx
```diff
@@ -0,0 +1,81 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import DocumentTitle from 'react-document-title';
+
+const BLINKING_INTERVAL = 1500;
+
+export class MessageIndicatorComponent extends Component {
+    state = {
+        initialDocumentTitle: global.document ? document.title : '',
```
mspensieri: Should this value be refreshed at some point? If the user switches pages or the page title changes at some point, our value will be stale

lemieux: Good point, but I wonder how we can effectively track that. I mean, what if the page change occurs while the message is shown... I guess the user's app will update the title when changing url...?

mspensieri: I think before every change we should first check if the title is what we expect, something like this:

```
onTick() {
  if(title === "{name} said"){
    title = this.initialTitle;
  }else if(title == this.initialTitle){
    title = '{name} said';
  }else{
    this.initialTitle = title;
    title = '{name} said';
  }
}
```

lemieux: yeah good idea, fix incoming

src/js/components/message-indicator.jsx
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    blinkTitle() {
+        if (!this.blinkInterval) {
+            const fn = () => {
+                const {unreadCount, messages, messageIndicatorTitle} = this.props;
+                const {currentTitle} = this.state;
+                let {initialDocumentTitle} = this.state;
+                
+                if (document.title !== initialDocumentTitle && !document.title.endsWith(messageIndicatorTitle.replace('{name}', ''))) {
```
lemieux: @mspensieri like this?

mspensieri: Yep. Why `endsWith` though?

lemieux: fastest way to check if it's using the text in the strings list without having to go through the conversation to fetch the last appMaker's name.

mspensieri: Don't think that check works if the guy customized his string though

```
"User Mike replied".endsWith("User {name} replied".replace('{name}', ''))

=> false
```

lemieux: true :(

lemieux: I'll add a reverse loop to start from the end of the conversation and get that name

mspensieri: What if we just stored the last set "smooch" title in the state?

README.md
```diff
@@ -99,7 +99,12 @@ var skPromise = Smooch.init({
         settingsInputPlaceholder: 'Your email address',
         settingsSaveButtonText: 'Save',
         settingsHeaderText: 'Email Settings',
-        settingsNotificationText: 'In case we\'re slow to respond you can <a href="#" data-ui-settings-link>leave us your email</a>.'
+        settingsNotificationText: 'In case we\'re slow to respond you can <a href data-ui-settings-link>leave us your email</a>.',
+        actionPaymentError: 'An error occurred while processing the card. <br> Please try again or use a different card.',
+        actionPaymentCompleted: 'Payment Completed',
+        messageError: 'An error occured while sending your message. Please try again.',
+        invalidFileError: 'Sorry, but only images are supported currently. Please choose a file with a supported extension (jpg, jpeg, png, gif, or bmp).',
+        messageIndicatorTitle: '{name} said...'
```
jpjoyal: Let's go with "(1) New message" & (2) "New messages" as name is will generate a 🤔  face most of the time for the user. Its very likely that its the first time you see this name

lemieux: done.

src/js/components/widget.jsx
```diff
@@ -10,16 +10,17 @@ import { Settings } from 'components/settings';
 import { Notification } from 'components/notification';
 import { ErrorNotification } from 'components/error-notification';
 import { ChatInput } from 'components/chat-input';
+import { MessageIndicator } from 'components/message-indicator';
```
Mario54: nit, call this `NewMessageIndicator`?

Mario54:
👍

mspensieri:
:+1:

PR #234, opened 2016-04-12
New message sound notification

lemieux:
Second part of new message notification : sound.

This would play a sound notification when the a new message comes in and the page in now in focus.

I also improve the way we clear the unread notification. It now clears whenever you click, mousewheel, or touch the widget.

@mspensieri @dannytranlx @Mario54


mspensieri:
:+1: The mp3 will be hosted on cdn right?

lemieux:
Yup, that reminds me to add the file to the uploaded file list...

dannytranlx:
Do we expose the file path? It would be nice to let the users use their own sound file 😂

lemieux:
Nope, and, nope.

PR #235, opened 2016-04-13
Add postback action support

dannytranlx:


src/js/components/action.jsx
```diff
@@ -112,6 +117,15 @@ export class ActionComponent extends Component {
                     </div>
                     );
             }
+        } else if (this.props.type === 'postback') {
+            return (
+                <div className='sk-action'>
+                    <button className='btn btn-sk-primary'
```
lemieux: don't forget the style

lemieux:
You'll need to publish smooch-core first and bump up version here before merging. Otherwise :+1:

src/js/components/action.jsx
```diff
@@ -112,6 +117,15 @@ export class ActionComponent extends Component {
...
+                            onClick={ this.onPostbackClick.bind(this) }>
```
Mario54: I might be better to inline that function since it's not doing much:
`onClick={() => postPostback(this.props._id)}`

src/js/services/conversation-service.js
```diff
@@ -163,3 +163,10 @@ export function handleConversationUpdated() {
 
     return Promise.resolve();
 }
+
+export function postPostback(actionId) {
+    const {user} = store.getState();
+    return core().conversations.postPostback(user._id, actionId).then((response) => {
```
Mario54: I think the `.then` is not needed here

src/js/components/action.jsx
lemieux: I don't mind

Mario54:
👍

src/js/components/action.jsx
dannytranlx: yeah, wasn't sure what was the norm, will update!

Mario54: I know @lemieux doesn't like `.bind` though 😛

src/js/services/conversation-service.js
dannytranlx: why not

src/js/components/action.jsx
lemieux: yeah, if you're using `.bind` don't do it there.

lemieux: either do 
```
class Something {
   onPostbackClick = () => {...}
}
```
or do it in the constructor

src/js/services/conversation-service.js
lemieux: because `postPostback` already resolve with the response, you're adding a useless layer

src/js/components/action.jsx
dannytranlx: added

src/js/services/conversation-service.js
dannytranlx: ah sure!

src/js/components/action.jsx
dannytranlx: oh i'll update to that thing, because i am now handling states, so it needs its own method and scope!

src/js/components/action.jsx
```diff
@@ -17,6 +18,18 @@ export class ActionComponent extends Component {
         };
     }
 
+    onPostbackClick = () => {
+        this.setState({
+            state: 'processing'
+        });
+
+        postPostback(this.props._id).then(() => {
```
Mario54: Probably need a `catch` here to at least set the state to `done`, but we should do better error handling

Mario54: What do we do for stripe errors?

dannytranlx: We just catch and set to `offered` again!

Mario54: Actually, we do more than that, we show an error notification as well: https://github.com/smooch/smooch-js/blob/f/postback/src/js/services/stripe-service.js#L8

src/js/components/action.jsx
```diff
@@ -17,6 +18,23 @@ export class ActionComponent extends Component {
...
+            this.setState({
+                state: 'paid'
```
Mario54: Should change those to be more general e.g. `done` and `normal` instead of `paid` and `offered`

dannytranlx: I was getting ready in case we were saving the state, those are the one in the `enum`. *shrug Sorry

src/js/components/action.jsx
dannytranlx: Interesting, I'll look into adding that notification too!

src/js/components/action.jsx
lemieux: I'd keep it like that, different actions can have different lifecycles.

lemieux: Actually we should have specific states for the postback action if we do track state.

Mario54: Well here the state isn't even used for the postback buttons, right?

mspensieri:
:+1:

Mario54:
👍

src/js/reducers/ui-reducer.js
```diff
@@ -15,6 +15,7 @@ const INITIAL_STATE = {
         settingsNotificationText: 'In case we\'re slow to respond you can <a href data-ui-settings-link>leave us your email</a>.',
         actionPaymentError: 'An error occurred while processing the card. <br> Please try again or use a different card.',
         actionPaymentCompleted: 'Payment Completed',
+        actionPostbackError: 'An error occured while processing your action. Please try again.',
```
jugarrit: occured > occurred

PR #242, opened 2016-04-26
Release 2.5.0

dannytranlx:
# New Stuff
- Added support for postback buttons, which can be used in automation flows. They will appear as regular buttons, but instead of redirecting to a url or showing a payment screen, it will post a webhook to your server when the user clicks one. More information [in the docs](http://docs.smooch.io/javascript/#postbacks)


Mario54:
👍

PR #246, opened 2016-05-13
When closing widget on mobile, prevent scrolling to the top of page

alavers:
@lemieux @jugarrit @jpjoyal 

The fix to make chat scrolling work on mobile also introduced a separate bug. Setting the html body width and height to 100% scrolls the page back up to the top, frustrating if you open the widget on a long page you've scrolled deeply into.

### DIFF
--- a/src/stylesheets/main.less
+++ b/src/stylesheets/main.less
@@ -69,7 +69,7 @@
     }
     @media (max-width: @screen-sm-min) {
         width: 100%;
-        height: 100%;
+        height: auto;
     }
 }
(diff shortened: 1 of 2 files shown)


lemieux:
:+1:

jpjoyal:
Looks good but worth testing on different devices / web sites

alavers:
- [ ] Chrome
- [ ] iOS
- [ ] Android
- [ ] Bootstrap (with/without)
- [ ] Embedded mode

PR #247, opened 2016-05-13
Release 2.5.1

alavers:
# Fixes
- Log warning when destroy is called before init
- When closing widget on mobile, prevent page from scrolling to the top
- Set widget height to auto to resolve an issue observed on some pages


lemieux:
:+1: do it!

PR #248, opened 2016-05-16
Use accent color for loading image animation

Mario54:
@lemieux @mspensieri @chloepouprom @liyefei737 @dannytranlx 

The loading animation was using the default blue color even if the accent color was set in the web app. With this change the accent color will be used if there is one.

web-loading-image

### DIFF
--- a/src/js/components/image-loading.jsx
+++ b/src/js/components/image-loading.jsx
@@ -1,10 +1,10 @@
 import React from 'react';
 
-export function ImageLoading({accentColor} = {}) {
+export function ImageLoading({color} = {}) {
     const bounceStyle = {};
 
-    if (accentColor) {
-        bounceStyle.backgroundColor = `#${accentColor}`;
+    if (color) {
+        bounceStyle.backgroundColor = `#${color}`;
     }
 
     return <div className='image-overlay'>
--- a/src/js/components/image-loading.jsx
+++ b/src/js/components/image-loading.jsx
@@ -1,11 +1,20 @@
 import React from 'react';
 
-export function ImageLoading() {
+export function ImageLoading({accentColor} = {}) {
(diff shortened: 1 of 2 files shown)


src/js/components/image-loading.jsx
```diff
@@ -1,11 +1,20 @@
...
+    if (accentColor) {
+        bounceStyle.backgroundColor = `#${accentColor}`;
+    }
+
     return <div className='image-overlay'>
                <div className='three-bounce spinner'>
-                   <div className='bounce1'></div>
-                   <div className='bounce2'></div>
-                   <div className='bounce3'></div>
+                   <div className='bounce1'
```
Mario54: We could refactor this part to remove the duplication, but I'm not sure it's worth the trouble?

mspensieri:
👍

chloepouprom:
👍

src/js/components/image-loading.jsx
lemieux: /shrug you're already in the image-loading component to abstract that kind of things... not sure how "optimal" it's worth going.

Mario54: Yeah, it's fine as-is 😛

src/js/components/image-message.jsx
```diff
@@ -24,7 +24,7 @@ export class ImageMessage extends Component {
     render() {
         const preloader = () => {
             return <div className='preloader-container'>
-                       <ImageLoading />
+                       <ImageLoading accentColor={ this.props.accentColor } />
```
lemieux: just call the prop `color` ... it defines the color of the whole component

lemieux:
:+1:

jugarrit:
Purty

PR #249, opened 2016-05-17
Automatically dismiss error notifications after 10 seconds

Mario54:
@lemieux @chloepouprom @liyefei737 @alavers @dannytranlx 

Using `redux-thunk` the `showErrorNotification` action will also dispatch a `hideErrorNotification` action after 10 seconds.

The style of the "x" button was changed, so that it is always shown to the user (not only when hovering the notification).

### DIFF
--- a/src/js/actions/app-state-actions.js
+++ b/src/js/actions/app-state-actions.js
@@ -118,7 +118,7 @@ export function setServerURL(url) {
 export function showErrorNotification(message) {
     return (dispatch) => {
         setTimeout(() => {
-            dispatch(hideErrorNotification(message));
+            dispatch(hideErrorNotification());
         }, 10000);
 
         dispatch({
@@ -128,10 +128,9 @@ export function showErrorNotification(message) {
     };
 }
 
-export function hideErrorNotification(message) {
+export function hideErrorNotification() {
     return {
-        type: HIDE_ERROR_NOTIFICATION,
-        message
+        type: HIDE_ERROR_NOTIFICATION
     };
 }
 
--- a/src/js/actions/app-state-actions.js
+++ b/src/js/actions/app-state-actions.js
@@ -18,8 +18,6 @@ export const SET_EMBEDDED = 'SET_EMBEDDED';
 export const ENABLE_IMAGE_UPLOAD = 'ENABLE_IMAGE_UPLOAD';
 export const DISABLE_IMAGE_UPLOAD = 'DISABLE_IMAGE_UPLOAD';
 
-const NOTIFICATION_HIDE_DELAY = 5000;
-
 export function toggleWidget() {
     return {
         type: TOGGLE_WIDGET
@@ -99,14 +97,8 @@ export function unsetEmailReadonly() {
 }
 
 export function showSettingsNotification() {
-    return (dispatch) => {
-        dispatch({
-            type: SHOW_SETTINGS_NOTIFICATION
-        });
-
-        setTimeout(() => {
-            dispatch(hideSettingsNotification());
-        }, NOTIFICATION_HIDE_DELAY);
+    return {
+        type: SHOW_SETTINGS_NOTIFICATION
     };
 }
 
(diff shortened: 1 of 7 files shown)


src/js/actions/app-state-actions.js
```diff
@@ -116,9 +116,15 @@ export function setServerURL(url) {
 }
 
 export function showErrorNotification(message) {
-    return {
-        type: SHOW_ERROR_NOTIFICATION,
-        message
+    return (dispatch) => {
+        setTimeout(() => {
+            dispatch(hideErrorNotification(message));
```
Mario54: Here, it means any error notifications will be dismissed, even if it didn't originate in the same call to `showErrorNotification`.

We could send an id when dispatching the `SHOW_ERROR_NOTIFICATION` action which would have to be sent with its associated `HIDE_ERROR_NOTIFICATION` action to make sure the right notification is dismissed. Thoughts @lemieux?

lemieux: Doesn't really, matter, the widget only supports one notification at the same time anyway.

By looking at this, can you do this little piece of clean up : remove `message` from the action... I don't know why it has an argument for dismissing it... probably a bad copy paste.

lemieux:
:+1: once tests are fixed

Mario54:
@lemieux I introduced the [redux-mock-store](https://github.com/arnaudbenard/redux-mock-store) package in 8f1f305 to make it easier to test async actions.

test/specs/services/conversation-service.spec.js
```diff
@@ -474,7 +474,7 @@ describe('Conversation service', () => {
 
                 it('should show an error notification', () => {
                     return conversationService.uploadImage({}).catch(() => {
-                        mockedStore.dispatch.should.have.been.calledWith({
+                        mockedStore.getActions().should.include({
```
lemieux: nice!

lemieux:
Very nice! Much easier to test when you can have 2 or 3 actions dispatched for a single thing.

src/js/actions/app-state-actions.js
Mario54: Fixed in 1d26a18

test/utils/redux.js
```diff
@@ -12,10 +13,13 @@ function restoreAppStore() {
...
 
 export function createMockedStore(sinon, mockedState = {}) {
-    const store = createStore(() => mockedState);
+    const middlewares = [thunkMiddleware];
+    const mockStore = configureStore(middlewares);
+    const store = mockStore(mockedState);
 
     sinon.spy(store, 'dispatch');
     store.restore = restoreAppStore;
+    store.subscribe = () => Function();
```
liyefei737: subscribe to empty function? or maybe I don't know enough about redux yet lol...

lemieux: `Function` with a capital F?

Mario54: @liyefei737 This parts of the code creates a mocked store, meaning that it will fake the actual store and provide an implementation that's easier to test.

In this case, I changed the `subscribe` function to make it return an empty function because `redux-mock-store` was making it return `null`.

Maybe the code is confusing (perhaps I should change that part)

```js
store.subscribe = () => Function();
// is equivalent to...
store.subscribe = function() {
    return function() {};
};
```

lemieux: @liyefei737 it doesn't subscribe to an empty function, it replaces `subscribe` with an empty one. I'd actually do `store.subscribe = sinon.stub()` instead in case we need to test it somehow.

Mario54: @lemieux See above. Invoking `Function` like this `Function()` will just return an empty function.

I'll change it so `Function()` is not used anymore

lemieux: yeah do `store.subscribe = sinon.stub().returns(Function())` instead.

Mario54: Good point @lemieux

chloepouprom:
👍

dannytranlx:
👍

PR #250, opened 2016-05-17
Handle multiple image upload

chloepouprom:
When multiple images were uploading at the same time, some would get hidden until upload was successful. Refer to https://trello.com/c/Ouf8Hy8L/3212-web-sdk-when-uploading-multiple-images-at-the-same-time-some-are-hidden-until-upload-is-successful for example.

@lemieux @Mario54 @alavers @liyefei737

### DIFF
--- a/src/js/services/conversation-service.js
+++ b/src/js/services/conversation-service.js
@@ -8,7 +8,6 @@ import { immediateUpdate } from 'services/user-service';
 import { initFaye } from 'utils/faye';
 import { observable } from 'utils/events';
 import { resizeImage, getBlobFromDataUrl, isFileTypeSupported, mergeMessages } from 'utils/media';
-const _ = require('underscore');
 
 export function handleFirstUserMessage(response) {
     const state = store.getState();
--- a/src/js/services/conversation-service.js
+++ b/src/js/services/conversation-service.js
@@ -7,7 +7,7 @@ import { core } from 'services/core';
 import { immediateUpdate } from 'services/user-service';
 import { initFaye } from 'utils/faye';
 import { observable } from 'utils/events';
-import { resizeImage, getBlobFromDataUrl, isFileTypeSupported, mergeMessages } from 'utils/media';
+import { resizeImage, getBlobFromDataUrl, isFileTypeSupported } from 'utils/media';
 
 export function handleFirstUserMessage(response) {
     const state = store.getState();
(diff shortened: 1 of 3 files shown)


src/js/services/conversation-service.js
```diff
@@ -95,7 +96,9 @@ export function uploadImage(file) {
             return core().conversations.uploadImage(user._id, blob, {
                 role: 'appUser'
             }).then((response) => {
-                store.dispatch(setConversation(response.conversation));
+                // Merge messages from response with messages from store state
+                const conversation = mergeMessages(response, store.getState().conversation.messages, message);
```
lemieux: I would move that logic in the reducer. That way it would get executed everytime it gets dispatched,

src/js/utils/media.js
```diff
@@ -1,4 +1,5 @@
 import loadImage from 'lib/load-image';
+const _ = require('underscore');
```
lemieux: Hmm I'm pretty sure we only use underscore in tests but not in the actual lib code. We do it that way to not add a dependency that could be replaced by vanilla js and reduce the bundle size.

src/js/utils/media.js
```diff
@@ -57,3 +58,16 @@ export function resizeImage(file) {
         return canvas.toDataURL('image/jpeg');
     });
 }
+
+export function mergeMessages(response, storeMessages, message) {
```
lemieux: I'd put that directly in the reducer file too since it's really specific to that part of the lib. Also the media utils is more about images :)

src/js/utils/media.js
```diff
@@ -57,3 +58,16 @@ export function resizeImage(file) {
...
+    const responseMessages = response.conversation.messages;
+    const messages = _.filter(
+        _.unique(
```
lemieux: did you experience a case where you had double messages? the server should return message in double

lemieux: Ahhh nevermind, I see what you did there, you need unique because of the union

src/js/services/conversation-service.js
```diff
@@ -7,7 +7,8 @@ import { core } from 'services/core';
 import { immediateUpdate } from 'services/user-service';
 import { initFaye } from 'utils/faye';
 import { observable } from 'utils/events';
-import { resizeImage, getBlobFromDataUrl, isFileTypeSupported } from 'utils/media';
+import { resizeImage, getBlobFromDataUrl, isFileTypeSupported, mergeMessages } from 'utils/media';
+const _ = require('underscore');
```
liyefei737: _ not used in this file? rest 👍

PR #251, opened 2016-05-19
Fix disappearing image

chloepouprom:
@lemieux @liyefei737 @Mario54 @alavers @jugarrit 

This fixes the multiple image upload bug from https://trello.com/c/Ouf8Hy8L/3212-web-sdk-when-uploading-multiple-images-at-the-same-time-some-are-hidden-until-upload-is-successful

- A REPLACE_MESSAGE action is used instead.
- The SET_CONVERSATION action now merges messages from the store and from the response to ensure that no messages are lost.
- Unit tests were added for the Conversation Reducer.

### DIFF
--- a/src/js/reducers/conversation-reducer.js
+++ b/src/js/reducers/conversation-reducer.js
@@ -6,22 +6,22 @@ const INITIAL_STATE = {
     unreadCount: 0
 };
 
-const sortMessages = (messages) => messages.sort((a, b) => {
+const sortMessages = (messages) => messages.sort((messageA, messageB) => {
     // received is undefined when it's the temp message from the user
-    if (!a.received && !b.received) {
+    if (!messageA.received && !messageB.received) {
         // `_tempSent` is a local only prop
-        return a._tempSent - b._tempSent;
+        return messageA._tempSent - messageB._tempSent;
     }
 
-    if (!a.received) {
+    if (!messageA.received) {
         return 1;
     }
 
-    if (!b.received) {
+    if (!messageB.received) {
         return -1;
     }
 
-    return a.received - b.received;
+    return messageA.received - messageB.received;
 });
 
 const addMessage = (messages, message) => {
@@ -47,18 +47,18 @@ const replaceMessage = (messages, query, newMessage) => {
     return [...messages.slice(0, index), newMessage, ...messages.slice(index + 1)];
 };
 
-const isEqual = (a, b) => {
-    if (a._id && b._id && a._id === b._id) {
+const isEqual = (messageA, messageB) => {
+    if (messageA._id && messageB._id && messageA._id === messageB._id) {
         return true;
     }
 
-    if (!a._id || !b._id) {
-        if (a.role === b.role) {
-            if (a.text && b.text && a.text === b.text) {
+    if (!messageA._id || !messageB._id) {
+        if (messageA.role === messageB.role) {
+            if (messageA.text && messageB.text && messageA.text === messageB.text) {
                 return true;
             }
 
-            if (a.mediaType === b.mediaType && a.mediaUrl === b.mediaUrl) {
+            if (messageA.mediaType === messageB.mediaType && messageA.mediaUrl === messageB.mediaUrl) {
                 return true;
             }
         }
@@ -67,14 +67,14 @@ const isEqual = (a, b) => {
     return false;
 };
 
-const mergeMessages = (a, b) => {
+const mergeMessages = (messagesA, messagesB) => {
     // concat will make a union out of both arrays
-    return removeDuplicates(a.concat(b));
+    return removeDuplicates(messagesA.concat(messagesB));
 };
 
-const removeDuplicates = (array) => {
+const removeDuplicates = (messages) => {
     // reduce will return a new array, the function used in the reduction strips out duplicates
-    return array.reduce((filteredMessages, nextMessage) => {
+    return messages.reduce((filteredMessages, nextMessage) => {
         const message = filteredMessages.find((m) => isEqual(m, nextMessage));
         if (message) {
             return filteredMessages;
--- a/src/js/reducers/conversation-reducer.js
+++ b/src/js/reducers/conversation-reducer.js
@@ -48,11 +48,11 @@ const replaceMessage = (messages, query, newMessage) => {
 };
 
 const isEqual = (messageA, messageB) => {
-    if (messageA._id && messageB._id && messageA._id === messageB._id) {
+    if (messageA._id && messageB._tempId && messageA._id === messageB._tempId) {
         return true;
     }
 
-    if (!messageA._id || !messageB._id) {
+    if (!messageA._id || !messageB._tempId) {
         if (messageA.role === messageB.role) {
             if (messageA.text && messageB.text && messageA.text === messageB.text) {
                 return true;
--- a/src/js/reducers/conversation-reducer.js
+++ b/src/js/reducers/conversation-reducer.js
@@ -69,8 +69,12 @@ const isEqual = (a, b) => {
 
 const mergeMessages = (a, b) => {
     // concat will make a union out of both arrays
+    return removeDuplicates(a.concat(b));
+};
+
+const removeDuplicates = (array) => {
     // reduce will return a new array, the function used in the reduction strips out duplicates
-    return a.concat(b).reduce((filteredMessages, nextMessage) => {
+    return array.reduce((filteredMessages, nextMessage) => {
         const message = filteredMessages.find((m) => isEqual(m, nextMessage));
         if (message) {
             return filteredMessages;
@@ -78,7 +82,6 @@ const mergeMessages = (a, b) => {
         return [...filteredMessages, nextMessage];
     }, []);
 };
-
 
 export function ConversationReducer(state = INITIAL_STATE, action) {
     switch (action.type) {
--- a/src/js/reducers/conversation-reducer.js
+++ b/src/js/reducers/conversation-reducer.js
@@ -44,15 +44,7 @@ const replaceMessage = (messages, query, newMessage) => {
     }
 
     const index = messages.indexOf(existingMessage);
-
-    // get rid of duplicate values
-    return [...messages.slice(0, index), newMessage, ...messages.slice(index + 1)].reduce((filteredMessages, nextMessage) => {
-        const message = filteredMessages.find((m) => isEqual(m, nextMessage));
-        if (message) {
-            return filteredMessages;
-        }
-        return [...filteredMessages, nextMessage];
-    }, []);
+    return [...messages.slice(0, index), newMessage, ...messages.slice(index + 1)];
 };
 
 const isEqual = (a, b) => {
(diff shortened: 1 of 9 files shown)


src/js/services/conversation-service.js
```diff
@@ -43,12 +43,11 @@ export function sendChain(sendFn) {
...
     return sendChain(() => {
-        // add an id just to please React
-        // this message will be replaced by the real one on the server response
         const message = {
-            _id: Math.random(),
             role: 'appUser',
-            text
+            text,
+            _tempId: Math.random(),
+            sent: new Date()
```
lemieux: prefix `sent` with an underscore so we know it's local

lemieux: which makes me think that `_id` is confusing because it's not local... idk what the convention should be... nothing? double-underscores?

lemieux:
I'm wondering if all these array manipulations will be efficient with a long conversation :/

lemieux:
I can't really :+1: since I wrote half of it, but I approve the tests :)

src/js/services/conversation-service.js
```diff
@@ -95,13 +97,16 @@ export function uploadImage(file) {
...
             }).then((response) => {
-                store.dispatch(setConversation(response.conversation));
+                store.dispatch(replaceMessage({
+                    _tempId: message._tempId
+                }, response.message));
                 observable.trigger('message:sent', response.message);
                 return response;
-            }).catch(() => {
+            }).catch((err) => {
+                console.log(err);
```
Mario54: Forgot to remove this console.log?

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,80 @@ const addMessage = (messages, message) => {
...
+const replaceMessage = (messages, query, newMessage) => {
+    const existingMessage = messages.find((message) => matchMessage(message, query));
+    if (!existingMessage) {
+        return messages;
+    }
+
+    const index = messages.indexOf(existingMessage);
+
+    // get rid of duplicate values
+    return [...messages.slice(0, index), newMessage, ...messages.slice(index + 1)].reduce((filteredMessages, nextMessage) => {
```
Mario54: This line is a bit long, putting `.reduce(...)` on the next line would improve readability

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,80 @@ const addMessage = (messages, message) => {
...
+
+const replaceMessage = (messages, query, newMessage) => {
+    const existingMessage = messages.find((message) => matchMessage(message, query));
+    if (!existingMessage) {
+        return messages;
+    }
+
+    const index = messages.indexOf(existingMessage);
+
+    // get rid of duplicate values
```
Mario54: In what case can there be a duplicate value?

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,80 @@ const addMessage = (messages, message) => {
...
+        }
+    }
+
+    return false;
+};
+
+const mergeMessages = (a, b) => {
+    // concat will make a union out of both arrays
+    // reduce will return a new array, the function used in the reduction strips out duplicates
+    return a.concat(b).reduce((filteredMessages, nextMessage) => {
```
Mario54: You can probably extract a `removeDuplicates` function since you duplicated the logic to remove duplicates (line 49) 😛

src/js/services/conversation-service.js
Mario54: That is a bit confusing. Or maybe call it `_tempSent` like `_tempId`?

src/js/reducers/conversation-reducer.js
chloepouprom: The duplicate values showed up after an image had uploaded

lemieux: the duplication removal logical should happen where you merge the two arrays, not where you replace a message

src/js/reducers/conversation-reducer.js
chloepouprom: Hmm, this is only ever used in the context of the `reduce` though. Might be confusing if it's moved outside? @lemieux

lemieux: I see what @Mario54 is saying, I just think that it shouldn't be necessary to remove duplicate when replacing a message. There's no reason duplicated entry should be there.

However, you could extract the duplication removal logic in its own function 
```
const removeDuplicates = (array) => {
    return array.reduce(....);
```
and then do `return removeDuplicates(a.concat(b))`

Mario54: You could do `removeDuplicates(a.concat(b));` and the body of removeDuplicates would be the `reduce(...)`

```js
function removeDuplicates(messages) {
    messages.reduce((filteredMessages, nextMessage) => {
        ...
    });
}
```

Mario54:
👍

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,75 @@ const addMessage = (messages, message) => {
...
+    const existingMessage = messages.find((message) => matchMessage(message, query));
+    if (!existingMessage) {
+        return messages;
+    }
+
+    const index = messages.indexOf(existingMessage);
+    return [...messages.slice(0, index), newMessage, ...messages.slice(index + 1)];
+};
+
+const isEqual = (a, b) => {
```
jugarrit: nit: could we call these messageA and messageB?

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,75 @@ const addMessage = (messages, message) => {
...
+            if (a.mediaType === b.mediaType && a.mediaUrl === b.mediaUrl) {
+                return true;
+            }
+        }
+    }
+
+    return false;
+};
+
+const mergeMessages = (a, b) => {
```
jugarrit: same here

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,75 @@ const addMessage = (messages, message) => {
...
+    // concat will make a union out of both arrays
+    return removeDuplicates(a.concat(b));
+};
+
+const removeDuplicates = (array) => {
```
jugarrit: Can we call this messages?

jugarrit:
Wow interns taking the tough bugs :P
👍

src/js/reducers/conversation-reducer.js
alavers: replaceMessage already checks if the message is there or not before it proceeds. Do we need this ~O(n^2) removeDuplicates check?

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,75 @@ const addMessage = (messages, message) => {
...
+
+    return false;
+};
+
+const mergeMessages = (messagesA, messagesB) => {
+    // concat will make a union out of both arrays
+    return removeDuplicates(messagesA.concat(messagesB));
+};
+
+const removeDuplicates = (messages) => {
```
alavers: `replaceMessage` already checks if the message is there before proceeding. Do we need this O(n^2) remove duplicates function?

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,75 @@ const addMessage = (messages, message) => {
...
-                messages: sortMessages(action.conversation.messages)
+                messages: sortMessages(mergeMessages(state.messages, action.conversation.messages))
             });
         case ConversationActions.ADD_MESSAGE:
             return Object.assign({}, state, {
                 messages: addMessage(state.messages, action.message)
             });
+        case ConversationActions.REPLACE_MESSAGE:
+            return Object.assign({}, state, {
+                messages: removeDuplicates(replaceMessage(state.messages, action.queryProps, action.message))
```
lemieux: why do we need to do `removeDuplicates` here? (just echoing what @alavers said on an outdated diff)

src/js/reducers/conversation-reducer.js
alavers: also if you're worried about the function's efficiency then implement an object hash function for messages. Eg concat the id, role, text media type together and hash it. Then your remove duplicates can be O(n)

lemieux: Yeah I'm a bit worried about the efficiency of all of this, maybe we should do that.

chloepouprom: mergeMessages gets called without replaceMessages during the SET_CONVERSATION action

src/js/reducers/conversation-reducer.js
chloepouprom: Image upload works in 3 steps:
- ADD_MESSAGE of the "sending" image [1]
- ADD_MESSAGE of the "received" image [2]
- REPLACE_MESSAGE of the "sending" by a different "received" [3]
This video makes it more clear: https://www.dropbox.com/s/vw6lcx86ddj0ri5/uplading_image.mov?dl=0

I added the removeDuplicates because I would end up with two copies of the "received" image.

Although, it might be worthwhile to re-visit the replaceMessage logic.

lemieux: why are we doing the second ADD_MESSAGE?

chloepouprom: Second ADD_MESSAGE is coming from here: https://github.com/smooch/smooch-js/blob/integration/src/js/utils/faye.js#L41

alavers: It might be simpler to change `addMessage` and have it look for any messages that need replacing, and just replace them then and there.

Then we just get rid of REPLACE_MESSAGE. Does that work @chloepouprom & @lemieux ?

chloepouprom:
I added all the changes as discussed @mspensieri @lemieux @alavers @jugarrit 
screen shot 2016-05-20 at 2 56 11 pm

src/js/reducers/conversation-reducer.js
```diff
@@ -36,22 +34,79 @@ const addMessage = (messages, message) => {
...
+};
+
+const mergeMessages = (messagesA, messagesB) => {
+    // concat will make a union out of both arrays
+    return removeDuplicates(messagesA.concat(messagesB));
+};
+
+const removeDuplicates = (messages) => {
+    let messagesNoDuplicates = [];
+    let messagesHash = {};
```
lemieux: I think you could just use an array here.

src/js/utils/faye.js
```diff
@@ -38,7 +39,9 @@ export function initFaye() {
         });
 
         return faye.subscribe(`/v1/conversations/${state.conversation._id}`, (message) => {
-            store.dispatch(addMessage(message));
+            if (!message.source.id || message.source.id !== getDeviceId()) {
```
lemieux: I don't think you need to check for `message.source.id` since it should be different than the device id anyway

lemieux:
now I'm :+1: this ! great job!

src/js/reducers/conversation-reducer.js
chloepouprom: Using an object will be more efficient (especially for larger conversations), since lookup doesn't require going through every element.

alavers: 👍

alavers:
👍

src/js/reducers/conversation-reducer.js
lemieux: then you probably don't need to assign a value to that key in the object /shrug

PR #257, opened 2016-05-25
Display messages sent by appMaker

chloepouprom:
This fixes the regression introduced by #253
- AppMaker messages are sent
- ConversationReducer was re-factored (namely, `isEqual` and `mergeMessages` were removed)
- More tests!

@lemieux @mspensieri @alavers @jugarrit @liyefei737 @Mario54

### DIFF
--- a/src/js/reducers/conversation-reducer.js
+++ b/src/js/reducers/conversation-reducer.js
@@ -48,11 +48,12 @@ const replaceMessage = (messages, query, newMessage) => {
 };
 
 const isEqual = (messageA, messageB) => {
+    // _tempId is a property of messages sent by the appUser
     if (messageA._id && messageB._tempId && messageA._id === messageB._tempId) {
         return true;
     }
 
-    if (!messageA._id || !messageB._tempId) {
+    if (!messageA._id && !messageB._id || !messageA._id && !messageB._tempId) {
         if (messageA.role === messageB.role) {
             if (messageA.text && messageB.text && messageA.text === messageB.text) {
                 return true;
--- a/src/js/reducers/conversation-reducer.js
+++ b/src/js/reducers/conversation-reducer.js
@@ -25,15 +25,8 @@ const sortMessages = (messages) => messages.sort((messageA, messageB) => {
 });
 
 const addMessage = (messages, message) => {
-    const existingMessage = messages.find((m) => isEqual(m, message));
-
-    if (existingMessage) {
-        return messages;
-    }
-
     return sortMessages([...messages, message]);
 };
-
 
 const matchMessage = (message, queryProps) => Object.keys(queryProps).every((key) => message[key] === queryProps[key]);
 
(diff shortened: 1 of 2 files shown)


test/specs/reducers/conversation-reducer.spec.js
```diff
@@ -18,14 +18,21 @@ const MESSAGE_2 = {
...
-const MESSAGE_TO_ADD = {
+const MESSAGE_FROM_APP_USER = {
     text: 'hey there!',
     role: 'appUser',
     received: 234.678,
     authorId: '8a9445dadad4862c2322db52',
     name: 'Calm Chimpanzee',
     _tempId: '123498001'
 };
+const MESSAGE_FROM_APP_MAKER = {
```
lemieux: can you also add tests for whispers while you are there?

src/js/reducers/conversation-reducer.js
```diff
@@ -48,11 +48,12 @@ const replaceMessage = (messages, query, newMessage) => {
 };
 
 const isEqual = (messageA, messageB) => {
+    // _tempId is a property of messages sent by the appUser
     if (messageA._id && messageB._tempId && messageA._id === messageB._tempId) {
         return true;
     }
 
-    if (!messageA._id || !messageB._tempId) {
+    if (!messageA._id && !messageB._id || !messageA._id && !messageB._tempId) {
```
lemieux: can you add a comment about this comparison? I know the context well and I have a hard time telling why we're comparing that way

lemieux: maybe just add a paragraph about the logic in this function, I don't get the `if` on line 52 either :(

test/specs/reducers/conversation-reducer.spec.js
```diff
@@ -105,13 +112,28 @@ describe('Conversation reducer', () => {
...
+                message: MESSAGE_FROM_APP_USER
             });
             afterState.messages.length.should.eq(3);
             afterState.messages[0].should.eql(MESSAGE_1);
-            afterState.messages[1].should.eql(MESSAGE_TO_ADD);
+            afterState.messages[1].should.eql(MESSAGE_FROM_APP_USER);
             afterState.messages[2].should.eql(UPLOADING_IMAGE);
         });
+
+        it('should add appMaker message', () => {
```
lemieux: Can you add these scenarios : multiple messages from the same appMaker and multiple messages from different appMakers ?

lemieux:
:+1: after more tests and comments

src/js/reducers/conversation-reducer.js
```diff
@@ -47,38 +40,12 @@ const replaceMessage = (messages, query, newMessage) => {
...
-
 const removeDuplicates = (messages) => {
-    let messagesNoDuplicates = [];
-    let messagesHash = {};
+    const messagesNoDuplicates = [];
+    const messagesHash = {};
 
     messages.forEach((message) => {
-        let key = message._id + message.role + message.mediaType;
+        const key = message._id + message.rosle + message.mediaType;
```
lemieux: `rosle` ?

src/js/reducers/conversation-reducer.js
```diff
@@ -25,16 +25,9 @@ const sortMessages = (messages) => messages.sort((messageA, messageB) => {
 });
 
 const addMessage = (messages, message) => {
-    const existingMessage = messages.find((m) => isEqual(m, message));
```
mspensieri: I guess this, and the associated methods, became useless after the check for `source.id` ?

mspensieri:
👍

src/js/reducers/conversation-reducer.js
lemieux: yeah we figured it was useless. they would be useful when calling `setConversation` but it's only used when the conversation hasn't started yet and has no existing messages.

PR #259, opened 2016-05-26
Fix image rendering when uploading multiple

lemieux:
A rendering bug was happening when uploading multiple images in row. Since we were using `index` as the key in the message list, React was rerendering the message in a weird way and the same image was reused multiple times until all uploads were finished.

I also renamed `_temp*` variable to `_client*` to make it clearer it's in the client.

@chloepouprom @mspensieri @dannytranlx @liyefei737 @Mario54

### DIFF
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -61,8 +61,8 @@ export class ConversationComponent extends Component {
     }
 
     render() {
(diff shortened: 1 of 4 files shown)


chloepouprom:
👍

Mario54:
👍

PR #261, opened 2016-05-26
Better packaging for npm

### LINKED ISSUE #245: Make React a peer dependency
As we're using React 15 this plugin breaks our build because it's requiring React 14.

According to [this answer](http://stackoverflow.com/a/30454133/2418739) you should make React a peer dependency. Could you please release a new version with this in mind?

lemieux:
Here's an overview of what's going on in this PR : 

- `react` and `react-dom` are now peerDependencies to allow people to use their existing version when using npm
- new steps in the build were added to generate a transpiled version of the source code instead of generating a big bundle for npm
- all assets references were moved to a single file (explanation below)
- a new webpack config was added to generate the asset file (explanation below)
- version information was moved in a file (explanation below)
- aliases were removed from webpack, relative path are now used (explanation below)
- stylesheet is now only injected on render and removed on destroy
- squeezed in an upgrade for babel to be compatible with node 6 and webpack

## Build process
Bower and CDN build process wasn't changed. It still relies on a bundled version of the code. The npm version is different. Instead of building a big bundle, the files are transpiled by babel into `lib/`. Then, we generate the assets file (src/utils/assets.js, see below) and replace the babelified one. We then tell package.json to use `lib/main.js` as the entry point.

## Assets
Assets were moved to a centralized file to allow the dev/bundled code to use webpack loaders and to have a separate webpack config to generate that file for npm. Basically, that config generates only one file (assets.js) and it replaces the babelified one (which only do the requires, the webpackified one actually has the loader code in it). See this gist to understand the difference between both files : https://gist.github.com/lemieux/ceb0aef885134945bbb374bc2c21e3f9

## Version
Previously the version was injected as a global by Webpack, which means you get undefined variables when using the transpiled version. Why not just `require('package.json').version ? Pretty much because in the bundle version it would include the whole package.json file just for one value. The version bump mechanism will take care of bumping that version during the deployment.

## Aliases and relative paths
Aliases rely on the fact that the code is built by webpack. When transpiling the code, the requires are still using the aliases which mean nothing in the context of a different webpack setup or when using browserify. Using relative paths fixes all of it and I removed the aliases to make sure we don't use them by accident and break the npm build.

This will fix #245. I tested the new packaging using the bundled file, in a browserify project and in a Webpack project (and included notes in the readme about webpack).

@alavers @chloepouprom @mspensieri @dannytranlx @liyefei737 @Mario54


lemieux:
This will also require a smooch-core upgrade to prevent duplication of babel-runtime 
https://github.com/smooch/smooch-core-js/pull/34#issuecomment-221999903

src/js/constants/version.js
```diff
@@ -0,0 +1 @@
+export const VERSION = '2.6.1';
```
mspensieri: we're up to 2.6.3 now

lemieux: I'll adjust on merge to make sure I have the latest.

Gruntfile.js
```diff
@@ -150,12 +150,22 @@ module.exports = function(grunt) {
                     ].join(' && ');
                 }
             },
+            addLib: {
+                cmd: function() {
+                    return [
+                        'git add --force lib/'
+                    ].join(' && ');
```
mspensieri: what's the join for? future-proofing?

lemieux: ehh, copy paste from above.. I guess I can remove both `join`

mspensieri:
LGTM 👍

package.json
```diff
@@ -134,5 +136,11 @@
     "webpack": "1.12.9",
     "webpack-cli": "1.0.0",
     "webpack-dev-server": "1.14.0"
+  },
+  "peerDependencies": {
+    "babel-polyfill": "6.9.0",
+    "babel-runtime": "6.9.0",
+    "react": "0.14.7",
+    "react-dom": "0.14.7"
```
Mario54: Shouldn't this be `0.14.x` or something more lenient?

From https://nodejs.org/en/blog/npm/peer-dependencies/

> One piece of advice: peer dependency requirements, unlike those for regular dependencies, should be lenient. You should not lock your peer dependencies down to specific patch versions. It would be really annoying if one Chai plugin peer-depended on Chai 1.4.1, while another depended on Chai 1.5.0, simply because the authors were lazy and didn't spend the time figuring out the actual minimum version of Chai they are compatible with.

> The best way to determine what your peer dependency requirements should be is to actually follow semver. Assume that only changes in the host package's major version will break your plugin. Thus, if you've worked with every 1.x version of the host package, use "~1.0" or "1.x" to express this. If you depend on features introduced in 1.5.2, use ">= 1.5.2 < 2".

test/specs/components/action.spec.jsx
```diff
@@ -3,18 +3,17 @@ import React from 'react';
...
+import { mockAppStore } from '../../utils/redux';
+import { mockComponent } from '../../utils/react';
+
+import { ActionComponent } from '../../../src/js/components/action.jsx';
+import { LoadingComponent } from '../../../src/js/components/loading.jsx';
 
-const stripeService = require('services/stripe-service');
-const userService = require('services/user-service');
-const conversationService = require('services/conversation-service');
+const stripeService = require('../../../src/js/services/stripe-service');
```
Mario54: Can we keep the alias for tests?

Mario54:
👍

test/specs/components/action.spec.jsx
lemieux: I'd rather not using aliases at all... we don't add new test files everyday for the widget.. it's mostly a write once thing

package.json
lemieux: true, I'll use that... if you are using npm3, peerDependencies just issue warnings anyway... this will be useful for npm2 users

PR #262, opened 2016-05-26
Set conversation if sent message is the first one

lemieux:
Since we weren't using set conversation anymore, the conversation id was never set in the store for a new user and it prevented faye to connect properly to the server.

This is urgent. I'll deploy ASAP.

@mspensieri @alavers @dannytranlx @Mario54


alavers:
👍

jugarrit:
👍

PR #264, opened 2016-05-26
Remove sound notifications on initial load

dannytranlx:
Sound notifications won't play on initial load anymore. Only on subsequent message received. 

The initial `hasFocus` was also set as `false`, so if you never `blur` or `focus` the window, it will assume the window to be not focused thus playing the sound. This will be fixed too. Grabbing the current state with `document.hasFocus()`. Not sure that is supported on most browser however.

I could also change the initial `hasFocus` to `true`. _Most_ of the time, we can assume the user _is_ on the page when loading the page. The only exception is if he `Cmd+Click` or `opened in new tab` the page (in which case, we won't play the sound anyway, because it will be on initial load).

@lemieux 
@chloepouprom 
@mspensieri 
@jugarrit

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -81,6 +81,7 @@ function onWindowBlur() {
 }
 
 export function monitorBrowserState() {
+    store.dispatch(hasFocus(document.hasFocus()));
     window.addEventListener('focus', onWindowFocus);
     window.addEventListener('blur', onWindowBlur);
 }
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -64,6 +64,7 @@ observable.on('message:received', (message) => {
 });
 
 let lastTriggeredMessageTimestamp = 0;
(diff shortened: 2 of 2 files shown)


src/js/utils/dom.js
```diff
@@ -81,6 +81,7 @@ function onWindowBlur() {
 }
 
 export function monitorBrowserState() {
+    store.dispatch(hasFocus(document.hasFocus()));
```
lemieux: compatibility seems alright for `document.hasFocus` on the main desktop browsers, but it's not clear on mobile (and opera doesn't support it). I would just do `document.hasFocus ? document.hasFocus() : true` or wrap it in a if and change the default reducer value to true.

lemieux:
:+1: this will need to be tested across the main browsers we support.

mspensieri:
👍

chloepouprom:
👍

PR #267, opened 2016-05-30
Update Web SDK UI

chloepouprom:
https://trello.com/c/eYxn0jf0/3456-0-5-web-ui-visual-update

With this PR:
- Header background color matches accentColor
- Messages are similar to SmoochBot (grouping, bubble styling)
- Width and height of widget varies according to size of screen
- Welcome message is centred

@lemieux @alavers @liyefei737 @Mario54 @mspensieri

### DIFF
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -62,31 +62,7 @@ export class ConversationComponent extends Component {
     }
 
     render() {
-        const processedMessages = [...this.props.conversation.messages];
-        let lastAuthor;
-        processedMessages.forEach((message, index) => {
-            const author = message.role === 'appUser' ? message.role : `${message.role}_${message.name}`;
-
-            if (!lastAuthor) {
-                lastAuthor = author;
-                message.firstInGroup = true;
-                message.lastInGroup = true;
-            }
-
-            if (lastAuthor === author) {
-                if (index > 0) {
-                    processedMessages[index - 1].lastInGroup = false;
-                    message.lastInGroup = true;
-                }
-            } else {
-                message.firstInGroup = true;
-                message.lastInGroup = true;
-            }
-
-            lastAuthor = author;
-        });
-
-        const messages = processedMessages.map((message) => {
+        const messages = [...this.props.conversation.messages].map((message) => {
 
             return <MessageComponent key={ message._clientId || message._id }
                                      accentColor={ this.props.settings.accentColor }
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -62,7 +62,7 @@ export class ConversationComponent extends Component {
     }
 
     render() {
-        const messages = [...this.props.conversation.messages].map((message) => {
+        const messages = this.props.conversation.messages.map((message) => {
 
             return <MessageComponent key={ message._clientId || message._id }
                                      accentColor={ this.props.settings.accentColor }
@@ -92,9 +92,9 @@ export class ConversationComponent extends Component {
                          style={ logoStyle }>
                         <a href='https://smooch.io/live-web-chat/?utm_source=widget'
                            target='_blank'><span>Messaging by</span> <img className='sk-image'
-                                                                                                                      src={ logo }
-                                                                                                                      srcSet={ `${logo} 1x, ${logo2x} 2x` }
-                                                                                                                      alt='smooch.io' /></a>
+                                                                                                                                    src={ logo }
+                                                                                                                                    srcSet={ `${logo} 1x, ${logo2x} 2x` }
+                                                                                                                                    alt='smooch.io' /></a>
                     </div>
                 </div>
             </div>
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -62,7 +62,32 @@ export class ConversationComponent extends Component {
     }
 
     render() {
-        const messages = this.props.conversation.messages.map((message) => {
+        const processedMessages = [...this.props.conversation.messages];
+        let lastAuthor;
+        processedMessages.forEach((message, index) => {
+            const author = message.role === 'appUser' ? message.role : `${message.role}_${message.name}`;
+
+            if (!lastAuthor) {
+                lastAuthor = author;
+                message.firstInGroup = true;
+                message.lastInGroup = true;
+            }
+
+            if (lastAuthor === author) {
+                if (index > 0) {
+                    processedMessages[index - 1].lastInGroup = false;
+                    message.lastInGroup = true;
+                }
+            } else {
+                message.firstInGroup = true;
+                message.lastInGroup = true;
+            }
+
+            lastAuthor = author;
+        });
+
+        const messages = processedMessages.map((message) => {
+
             return <MessageComponent key={ message._clientId || message._id }
                                      accentColor={ this.props.settings.accentColor }
                                      linkColor={ this.props.settings.linkColor }
--- a/src/js/components/message.jsx
+++ b/src/js/components/message.jsx
@@ -3,12 +3,32 @@ import React, { Component } from 'react';
 import { TextMessage } from './text-message';
 import { ImageMessage } from './image-message';
 import { ActionComponent } from './action';
-
+import { findDOMNode } from 'react-dom';
+import { getElementProperties } from '../utils/dom';
 
 export class MessageComponent extends Component {
     static defaultProps = {
         actions: []
     };
+
+    componentDidMount() {
+        if (this.props.actions.length === 0) {
+            this._restyleBubble();
+        }
+    }
+
+    _restyleBubble() {
+        const bubble = findDOMNode(this.refs.messageContent);
+        if (bubble) {
+            const messageElement = bubble.firstChild;
+            const messageProperties = getElementProperties(messageElement);
+            const bubbleProperties = getElementProperties(bubble);
+            const multiLineCheck = parseInt(bubbleProperties.fontSize) * 2;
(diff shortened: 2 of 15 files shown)


src/js/components/message.jsx
```diff
@@ -23,6 +23,7 @@ export class MessageComponent extends Component {
             <img className='sk-msg-avatar'
                  src={ this.props.avatarUrl } />
             );
+        const avatarPlaceHolder = isAppUser? null : (<div className='sk-msg-avatar-placeholder'/>);
```
lemieux: format your file! should have a space between `isAppUser` and `null`

lemieux:
Overall nice job, but I raised some concerns in Slack about couple of things. Also, can you update this branch with what is in integration so it can run on Node 6?

lemieux:
What happens if my accent color is too light? Can't read the header text?

src/js/components/conversation.jsx
```diff
@@ -62,7 +62,32 @@ export class ConversationComponent extends Component {
     }
 
     render() {
-        const messages = this.props.conversation.messages.map((message) => {
+        const processedMessages = [...this.props.conversation.messages];
+        let lastAuthor;
+        processedMessages.forEach((message, index) => {
+            const author = message.role === 'appUser' ? message.role : `${message.role}_${message.name}`;
```
mspensieri: This naming convention (`${message.role}_${message.name}`) would mean that whispers and replies from the same person will not be grouped, right? Maybe the prefix should just be `am_` or something

src/js/components/conversation.jsx
```diff
@@ -62,7 +62,32 @@ export class ConversationComponent extends Component {
...
+
+            if (!lastAuthor) {
+                lastAuthor = author;
+                message.firstInGroup = true;
```
mspensieri: I'm not sure if I like mutating the message objects like this. Maybe the algorithm could instead group messages into separate arrays, then use those arrays to power the render function instead of looking at the boolean flags?

src/js/components/conversation.jsx
```diff
@@ -62,7 +62,32 @@ export class ConversationComponent extends Component {
     }
 
     render() {
```
mspensieri: Does this two-pass processing of the messages cause any slowdown in rendering speed? Should the messages be pre-processed on change, instead of re-processed on each render?

src/js/components/message.jsx
```diff
@@ -35,16 +36,54 @@ export class MessageComponent extends Component {
         }
 
         const style = {};
-        if (isAppUser && this.props.accentColor) {
+
+        if (!this.props.mediaUrl) {
+          style.borderTopLeftRadius = style.borderTopRightRadius = style.borderBottomLeftRadius = style.borderBottomRightRadius = 16;
```
mspensieri: Can we extract the magic numbers `16` and `2` into constants at the top of the file?

src/js/components/message.jsx
```diff
@@ -35,16 +36,54 @@ export class MessageComponent extends Component {
...
+          if (isAppUser && this.props.accentColor) {
             style.backgroundColor = style.borderLeftColor = `#${this.props.accentColor}`;
+          }
+          if (this.props.firstInGroup && !this.props.lastInGroup) {
+            if (isAppUser) {
+              style.borderBottomRightRadius = 2;
```
mspensieri: Doing this styling inline means that these values can't be customized through CSS, right? @lemieux do you know if people ever want to customize the look of the bubbles? Should these be classes instead?

lemieux: Hmm right, instead of using inline style the widget should be using classes to allow people to override them.

src/js/components/message.jsx
lemieux: Use classes instead, make an array of them and use `join(' ')` when setting the className on the element. I suggest you create really small classes that targets only one thing and then compose the style with that.

src/js/components/conversation.jsx
lemieux: Commenting for both this and https://github.com/smooch/smooch-js/pull/267#discussion_r65170987. I think we should hold a second array in the conversation reducer with either a duplication of the message + those flags, or just the flags and you reconcile them in render. I think this double processing is killing performance in Smoochbot. What do you think @mspensieri ?

lemieux: That could even be an object instead with the message id as the key so you don't have to worry about sorting and the look would be quicker

mspensieri: Does that kind of stuff really belong in the reducer? It's not really data, it's more like markup used for presentation, so belongs more in the component, no? In fact that markup doesn't even really need to be persisted anywhere... during the `.map` loop [here](https://github.com/smooch/smooch-js/pull/267/files#diff-44044b6ef5eb6631027b51708c51c608R90), you can just look at the previous and next message to know how you should render the current one, there's no need to store those boolean flags anywhere other than in the props of the `MessageComponent`

Mario54: [reselect](https://github.com/reactjs/reselect) might be a good way to do this

lemieux: Yes you could do that, but then you need to reprocess the message array on each render. What I'm suggesting is to do that processing in the reducer when doing `ADD_MESSAGE` or `SET_CONVERSATION`. That way the processing happens only when it's really needed. The component can rerender when you don't really expect it (change of state somewhere, change a flag, etc.). 

And yes, it can belong in the reducer... you can pretty much put anything in the reducer. The reducer is also a good place to put state instead of using local component state.

src/js/components/conversation.jsx
chloepouprom: @mspensieri this logic is there to make sure messages from different appMakers won't be grouped together. What if I only use `message.name` when looking at appMakers? That way, messages from different appMakers will be separated, but whispers and messages from same appMaker will show up together.

chloepouprom:
Alright, made a few more changes @mspensieri @lemieux 
- Used 14px radius for messages 
- Added classes for the message groups and for the header (they are now customizable)
- Message grouping logic is in the conversation reducer.

lemieux:
works for me :+1:

src/js/components/conversation.jsx
mspensieri: Yep that works too

mspensieri:
👍

dannytranlx:
In any UI-related PRs, it's usually nice to have some handy screenshots :)

I've ran the branch locally, I am seeing a few issues. (I am on Chrome OSX)

* As an appMaker, I reply with an image and my avatar isn't aligned with the bottom of the image bubble.

image
* The appMaker name label is making the bubble larger than needed

image

* On a long message, word-wrap or something is creating weird padding
image

* Was this suppose to group?

image

* Buttons are missing side margins to match mock up and a bit of border radius

image

jugarrit:
Are multiple messages from the same person supposed to be 'merged'?
image
VS
image

chloepouprom:
Messages sent by the same appMaker will be merged together, but messages sent by different appMakers will not.
screen shot 2016-05-31 at 4 22 45 pm

Whispers and messages sent by same appMaker will be grouped together. Here, I check the name of the author to determine whether or not the messages should be grouped.
screen shot 2016-05-31 at 4 20 08 pm

jugarrit:
But the first ever message of the conversation was from me, not the app maker

jugarrit:
And that one isn't merging :P

jugarrit:
image

dannytranlx:
Whispers and appMakers are fine, look like I didn't pull the last commit you did about that!

jugarrit:
image

jugarrit:
No message merge love for images?

image

jugarrit:
Once the above concerns are fixed lgtm 👍

chloepouprom:
@dannytranlx "On a long message, word-wrap or something is creating weird padding"

That's because max-width is set to 200px. Not sure if there's much we can do about it.

dannytranlx:
Do we have to force to max-width? Hmm, not sure how we can handle that. We had the same issue on Android where the bubble width was calculated before the words were broken.

Expected result should be this:
image

lemieux:
I think there's nothing we can do about it.. I think I mess with that a lot in the past.. maybe you'll have more luck than me.

jugarrit:
@dannytranlx we did have the same issue on Android, and I don't think it's fixed!

src/stylesheets/conversation.less
```diff
@@ -210,6 +210,28 @@
 
             .sk-msg-image {
                 background-color: transparent;
+                &.sk-msg-appmaker-first {
```
jugarrit: Can you extract these into a mixin or something to avoid repetition?

src/stylesheets/conversation.less
```diff
@@ -188,6 +198,32 @@
                     color: white;
                 }
             }
+
+            .sk-msg-image {
+                background-color: transparent;
+                &.sk-msg-appmaker-first {
```
jugarrit: Can you extract these into a mixin or something to avoid repetition?

jugarrit: Also won't this only affect the right row?

src/js/components/conversation.jsx
```diff
@@ -62,7 +62,8 @@ export class ConversationComponent extends Component {
...
-        const messages = this.props.conversation.messages.map((message) => {
+        const messages = [...this.props.conversation.messages].map((message) => {
```
alavers: Cloning the array? Is it necessary?

alavers:
👍

PR #269, opened 2016-06-01
Fix width of message bubble

chloepouprom:
These changes modify the width of message bubbles (à la fb messenger). 

Before:

screen shot 2016-06-01 at 2 39 54 pm

After:
screen shot 2016-06-01 at 2 40 23 pm

Only small drawback is that the message width is calculated using `offsetWidth`, which can differ by 1px depending on the browser. Below is an example. If you guys think this could cause problems, let me know and I'll re-visit this.

Chrome (`width: 197px`):
screen shot 2016-06-01 at 2 43 54 pm

Firefox (`width: 196px`): 
screen shot 2016-06-01 at 2 43 45 pm

@dannytranlx @jugarrit @lemieux

### DIFF
--- a/src/js/components/message.jsx
+++ b/src/js/components/message.jsx
@@ -53,7 +53,7 @@ export class MessageComponent extends Component {
             <img className={ avatarClass.join(' ') }
                  src={ this.props.avatarUrl } />
             );
-        const avatarPlaceHolder = isAppUser ? null : (<div className='sk-msg-avatar-placeholder'/>);
+        const avatarPlaceHolder = isAppUser ? null : (<div className='sk-msg-avatar-placeholder' />);
 
         const message = this.props.mediaUrl ?
             <ImageMessage {...this.props} /> :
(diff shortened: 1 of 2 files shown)


jugarrit:
Sweet :)
👍

src/js/components/message.jsx
```diff
@@ -3,13 +3,42 @@ import React, { Component } from 'react';
 import { TextMessage } from './text-message';
 import { ImageMessage } from './image-message';
 import { ActionComponent } from './action';
+import { findDOMNode } from 'react-dom';
 
+const getElementProperties = (element) => {
```
lemieux: Can we move this to a utils instead?

lemieux:
:+1: don't worry about the 1px difference, it's mainly due to how each browser is rounding the values... 10000000% better than what we had before!

src/js/components/message.jsx
```diff
@@ -3,13 +3,33 @@ import React, { Component } from 'react';
...
+import { findDOMNode } from 'react-dom';
+import { getElementProperties } from '../utils/dom';
 
 export class MessageComponent extends Component {
     static defaultProps = {
         actions: []
     };
 
+    componentDidMount() {
+        if (this.props.actions.length == 0) {
```
mspensieri: `===`

src/js/utils/dom.js
```diff
@@ -91,3 +91,13 @@ export function stopMonitoringBrowserState() {
     window.removeEventListener('focus', onWindowFocus);
     window.removeEventListener('blur', onWindowBlur);
 }
+
+export function getElementProperties(element) {
+    return {
+        width: element.offsetWidth || 0,
+        height: element.offsetHeight || 0,
+        paddingLeft: window.getComputedStyle(element, null).getPropertyValue('padding-left'),
```
mspensieri: Should we cache the result of `getComputedStyle` in a var to avoid calling it 3 times?

mspensieri:
👍  otherwise

dannytranlx:
👍  beauty!

alavers:
I'm late to the party, but is there a reason this wasn't possible to do with CSS? Is it because of browser compat?

chloepouprom:
The width needs to be re-calculated after the message bubble has been rendered.

PR #277, opened 2016-06-08
Use media query breakpoints

chloepouprom:
Replace viewport attributes (`vh` and `vw`) with media query breakpoints.

- On large screens, widget will be 640x410.
- On smaller screens, widget will be 420x330.
- On phones and tablets, widget will take up width and height of device.

@lemieux @jugarrit @dannytranlx @mspensieri

### DIFF
--- a/src/stylesheets/animations.less
+++ b/src/stylesheets/animations.less
@@ -8,6 +8,11 @@
 
     -webkit-animation-fill-mode: forwards;
             animation-fill-mode: forwards;
+
+    @media (min-width: @screen-lg-min) {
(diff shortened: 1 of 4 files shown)


lemieux:
I pulled the branch and tested it, everything seems fine now but one thing : 

image

You might need to tweak the size in the embedded mode too, the settings wrapper isn't wide enough. Maybe it was an existing bug though. I don't think we need to have it 100% (because the text will wrap) but we could apply the same width you used for each breakpoints.

lemieux:
Also I don't get why the header is so big in embedded

lemieux:
:+1: works for me

PR #279, opened 2016-06-22
Fix Messenger button alignment

lemieux:
Since it was centered and the iframe width seems to change when the user is logged out, the button was misaligned because of the combination of align: center and the margin. By realigning to the left, it works great in both cases.

@Mario54 @dannytranlx @jugarrit @mspensieri @alavers


mspensieri:
👍

dannytranlx:
👍

alavers:
👍

PR #285, opened 2016-06-22
Fix Messenger plugin unsolicited re-rendering

dannytranlx:
On user clicks, the component was re-rendering, so it was killing the `iframe`making the request fail and the component flash.

@jpjoyal 
@lemieux 
@Mario54 
@jugarrit 
@chloepouprom 

https://trello.com/c/Awoob9x2/4017-web-sdk-linking-to-messenger-on-ios-often-takes-a-few-tries

### DIFF
--- a/src/js/components/channels/messenger-channel-content.jsx
+++ b/src/js/components/channels/messenger-channel-content.jsx
@@ -22,6 +22,9 @@ export class MessengerChannelContent extends Component {
         }
     };
 
+    shouldComponentUpdate(nextProps, nextState) {
+        return nextState.sdkBlocked;
+    }
 
     render() {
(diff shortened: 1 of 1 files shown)


mspensieri:
👍

src/js/components/channels/messenger-channel-content.jsx
```diff
@@ -22,6 +22,9 @@ export class MessengerChannelContent extends Component {
         }
     };
 
+    shouldComponentUpdate(nextProps, nextState) {
+        return nextState.sdkBlocked;
```
dannytranlx: This should be the only case we need re-rendering.

Mario54:
👍

jpjoyal:
glad we could fix this

PR #292, opened 2016-06-23
Bump up react-redux version

lemieux:
Some performance optimization was introduced in the newest version. I figure the widget was sometimes re-rendering itself for no reason and `connect` was doing a bad job of figuring out when to rerender.

@Mario54 @alavers @dannytranlx @jugarrit


dannytranlx:
Ho, maybe that would've fix the Facebook plugin issue!

👍

jugarrit:
👍

PR #295, opened 2016-06-27
Re-enable disabled ui tests 

chloepouprom:
This re-enables/fixes the UI tests that were previously disabled/failing. I also added a few more tests.

Still left to do:
- Fix tests for `faye` and for `conversation-service`. I will cover this in another PR.
- Add more tests and improve testing coverage. This is covered in another card.

@lemieux @jugarrit @alavers @mspensieri @liyefei737

### DIFF
--- a/test/specs/components/action.spec.jsx
+++ b/test/specs/components/action.spec.jsx
@@ -289,11 +289,12 @@ describe('Action', () => {
 
             beforeEach(() => {
 
-                appUtils.getIntegration.resolves({
-                    stripe: {
+                appUtils.getIntegration.resolves(
+                    {
+                        type: 'stripeConnect',
                         publicKey: 'key'
                     }
-                });
+                );
                 mockedStore = mockAppStore(sandbox, getStoreState({
                     user: {
                         email: ''
--- a/test/specs/components/action.spec.jsx
+++ b/test/specs/components/action.spec.jsx
@@ -89,7 +89,7 @@ describe('Action', () => {
         conversationService.postPostback.resolves();
 
         sandbox.stub(appUtils, 'getIntegration');
-        appUtils.getIntegration.resolves({});
+        appUtils.getIntegration.returns({});
     });
 
     afterEach(() => {
@@ -164,7 +164,7 @@ describe('Action', () => {
             });
 
             beforeEach(() => {
-                appUtils.getIntegration.resolves(
+                appUtils.getIntegration.returns(
                     {
                         type: 'stripeConnect',
                         publicKey: 'key'
@@ -213,7 +213,7 @@ describe('Action', () => {
             });
 
             beforeEach(() => {
-                appUtils.getIntegration.resolves(
+                appUtils.getIntegration.returns(
                     {
                         type: 'stripeConnect',
                         publicKey: 'key'
@@ -241,7 +241,7 @@ describe('Action', () => {
                 const props = getBuyProps();
 
                 beforeEach(() => {
-                    appUtils.getIntegration.resolves(
+                    appUtils.getIntegration.returns(
                         {
                             type: 'stripeConnect',
                             publicKey: 'key'
--- a/test/specs/components/action.spec.jsx
+++ b/test/specs/components/action.spec.jsx
@@ -243,6 +243,7 @@ describe('Action', () => {
         });
 
         it('should render a link', () => {
+            console.log(componentNode);
             const link = TestUtils.findRenderedDOMComponentWithTag(component, 'a');
             link.textContent = 'action text';
             link.href = 'fallback uri';
@@ -302,7 +303,7 @@ describe('Action', () => {
 
                 component = TestUtils.renderIntoDocument(<ParentComponentWithContext context={ context }
                                                                                      store={ mockedStore }
-                                                                                     accessElement='true'>
+                                                                                     accessElement={ true }>
                                                              <ActionComponent {...props} />
                                                          </ParentComponentWithContext>);
 
--- a/test/specs/components/action.spec.jsx
+++ b/test/specs/components/action.spec.jsx
@@ -302,7 +302,7 @@ describe('Action', () => {
 
                 component = TestUtils.renderIntoDocument(<ParentComponentWithContext context={ context }
                                                                                      store={ mockedStore }
-                                                                                     accessElement={ true }>
+                                                                                     withRef={ true }>
                                                              <ActionComponent {...props} />
                                                          </ParentComponentWithContext>);
 
@@ -339,7 +339,7 @@ describe('Action', () => {
 
                 component = TestUtils.renderIntoDocument(<ParentComponentWithContext context={ context }
                                                                                      store={ mockedStore }
-                                                                                     accessElement={ true }>
+                                                                                     withRef={ true }>
                                                              <ActionComponent {...props} />
                                                          </ParentComponentWithContext>);
 
@@ -376,7 +376,7 @@ describe('Action', () => {
 
                 component = TestUtils.renderIntoDocument(<ParentComponentWithContext context={ context }
                                                                                      store={ mockedStore }
-                                                                                     accessElement={ true }>
+                                                                                     withRef={ true }>
                                                              <ActionComponent {...props} />
                                                          </ParentComponentWithContext>);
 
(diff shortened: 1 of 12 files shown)


test/specs/components/action.spec.jsx
```diff
@@ -195,19 +221,24 @@ xdescribe('Action', () => {
         });
     });
 
-    describe('buy action without stripe keys', () => {
+    // Re-visit this test after fixing link fallback
```
lemieux: didn't @liyefei737 fix it?

test/utils/parent-component.jsx
```diff
@@ -18,12 +34,19 @@ export class ParentComponentWithContext extends Component {
     }
 
     render() {
+        // If we need to use functions on the child element, then add a 'ref' propType
+        const childElement = this.props.accessElement ? React.Children.map(
```
lemieux: `react-redux` uses `withRef` to tell if the the wrapped component should be accessible. I like that better than `accessElement` /shrug

test/utils/parent-component.jsx
```diff
@@ -1,10 +1,26 @@
...
+//                                          </ParentComponentWithContext>);
+// component.refs.childElement.someFunction();
+
 export class ParentComponentWithContext extends Component {
 
     static propTypes = {
-        context: PropTypes.object
+        context: PropTypes.object.required,
+        store: PropTypes.object.required,
+        accessElement: PropTypes.string
```
lemieux: shouldn't this be a bool6

test/utils/react.js
```diff
@@ -30,3 +30,13 @@ export function mockComponent(sinon, module, mockTagName = 'div', props = null)
...
 }
+
+export function getContext(context = {}) {
+    const defaultContext = {
+        app: {},
+        settings: {},
+        ui: {}
+    };
+
+    return Object.assign(defaultContext, context);
```
lemieux: death to object assign ;)

test/specs/components/action.spec.jsx
chloepouprom: We used to check for keys, but that got removed at some point: https://github.com/smooch/smooch-js/commit/e8bb52380f3ff23ff8fa589d21ee4157b2881c2d#diff-1176353b9d32bfd3b90e28ff3100b43cL90

test/utils/react.js
liyefei737: instead we should use ?

lemieux:
```
return  {
    ...defaultContext,
    ...context
}
```

note that you can only use that in the frontend, node doesn't support that.

alavers: Why death to Object.assign?

lemieux: I prefer by far the object spread syntax over object.assign. However, Object spread is still a proposal and it's available because we use a babel transform.

lemieux:
lgtm :+1:

test/specs/components/action.spec.jsx
```diff
@@ -94,6 +87,9 @@ xdescribe('Action', () => {
 
         sandbox.stub(conversationService, 'postPostback');
         conversationService.postPostback.resolves();
+
+        sandbox.stub(appUtils, 'getIntegration');
```
mspensieri: is it necessary to stub this? why can't we just set up the app with the proper integrations?

lemieux: I think it's a valid way to do it, we're not testing if `getIntegration` is returning the right value or not.. we want it to always return false for that test case. `getIntegration` should be tested on its own.

test/specs/components/action.spec.jsx
mspensieri: keys got moved in the init response, the check is still there

```
{
  publicKeys: {
    stripe: 'xyz'
  }
}
```

is now

```
{
   app : {
    integrations: [{
      type: 'stripeConnect',
      publicKey: 'xyz'
    }]
  }
}

File: test/specs/components/action.spec.jsx [hunk 6]
```diff
@@ -242,24 +278,41 @@ xdescribe('Action', () => {
...
+                    stripe: {
+                        appName: 'app-name',
+                        iconUrl: 'iconUrl'
+                    }
+                }
+            });
 
             beforeEach(() => {
 
+                appUtils.getIntegration.resolves({
```
mspensieri: Isn't `getIntegration` supposed to return something of the form

```
{
  type: 'stripeConnect',
  publicKey: 'abc'
}
```

?

test/specs/components/action.spec.jsx
```diff
@@ -271,20 +324,32 @@ xdescribe('Action', () => {
 
         describe('user has email', () => {
             const props = getBuyProps();
+            const context = getContext({
+                app: {
```
mspensieri: why do we have to set this up every time? All these tests can't possibly rely on `stripe`, do they?

test/specs/components/action.spec.jsx
chloepouprom: Yeah... `getIntegration` will be tested in the `appUtils` unit test. Although, there's not test for that yet...

test/specs/components/action.spec.jsx
```diff
@@ -195,19 +221,24 @@ xdescribe('Action', () => {
...
         const props = getBuyProps();
-
-        beforeEach(() => {
-            mockedStore = mockAppStore(sandbox, getStoreState({
-                app: {
-                    publicKeys: {
-                        stripe: false
-                    }
+        const context = getContext({
+            app: {
```
mspensieri: can you nest all the tests that require stripe under one `describe` and do this setup in a higher-level `beforeEach`?

test/specs/components/action.spec.jsx
lemieux: true

mspensieri: I feel like this is where mocking `getIntegration` hurts us. It's such a small function, all it does is filter an array.. why is it a util at all, and why does that deserve mocking?

mspensieri: can we validate input at the same time then? 

```
appUtils.getIntegration.withArgs('stripeConnect').returns(...)
```

or

```
appUtils.getIntegration.should.have.been.calledWith('stripeConnect')
```

test/specs/components/action.spec.jsx
```diff
@@ -156,290 +153,247 @@ describe('Action', () => {
...
-        beforeEach(() => {
-            mockedStore = mockAppStore(sandbox, getStoreState());
-            component = TestUtils.renderIntoDocument(<ParentComponentWithContext context={ context }
-                                                                                 store={ mockedStore }>
-                                                         <ActionComponent {...props} />
-                                                     </ParentComponentWithContext>);
-            componentNode = ReactDOM.findDOMNode(component);
-        });
+            beforeEach(() => {
+                appUtils.getIntegration.resolves(true);
```
lemieux: that should resolve with a real channel object

lemieux: and yeah based on @mspensieri it should probably test the input too

test/specs/components/action.spec.jsx
```diff
@@ -138,258 +144,318 @@ xdescribe('Action', () => {
...
 
-        it('should render a loading element', () => {
-            TestUtils.scryRenderedDOMComponentsWithClass(component, 'mockedStripe').length.should.be.eq(0);
-            TestUtils.scryRenderedDOMComponentsWithClass(component, 'mockedLoading').length.should.be.eq(1);
-            TestUtils.scryRenderedDOMComponentsWithClass(component, 'btn-sk-action-paid').length.should.be.eq(0);
-            componentNode.textContent.should.eq('');
-        });
-    });
+            beforeEach(() => {
+                appUtils.getIntegration.resolves(
```
mspensieri: Hmm actually It think what you want here is `returns`, no? `resolves` is supposed to return a promise that resolves to that value

mspensieri:
github is hiding it but I think this comment is still valid https://github.com/smooch/smooch-js/pull/295#discussion_r68630854

test/specs/components/action.spec.jsx
lemieux: should be `returns`

mspensieri: Why is the test still passing though 😕

lemieux: because it returns a truthty value (a promise) and in the component it's used like this 

```
if (this.props.type === 'buy' && stripeIntegration) {
```

then later it passes `stripeIntegration.publicKey` to the StripeCheckout component (which seems to don't care about it to be undefined, or it's actually mocked)

mspensieri:
👍

PR #296, opened 2016-06-28
Remove render patch for unwanted rerender

lemieux:
This is pretty much unneeded now since the react-redux upgrade fixed it.

I smoketested the widget with that modification and it went fine.

@dannytranlx @jugarrit @mspensieri @alavers @chloepouprom


dannytranlx:
👍

Mario54:
👍

PR #309, opened 2016-07-06
Use min and max variables for height breakpoints

lemieux:
At exactly 648px, the widget was half opened when it was supposed to be closed. That was caused by the overlap of two height breakpoints. This fixes this problem by having explicit max sizes depending on the next breakpoint.

@chloepouprom @alavers @dannytranlx @jugarrit @mspensieri

### DIFF
--- a/src/stylesheets/variables.less
+++ b/src/stylesheets/variables.less
@@ -49,9 +49,13 @@
 
 @font-size-base: 14px;
 
-@screen-lg-ht: 835px;
-@screen-md-ht: 648px;
-@screen-sm-ht: 488px;
+@screen-lg-ht-min: 835px;
+@screen-md-ht-min: 648px;
+@screen-sm-ht-min: 488px;
+
+@screen-md-ht-min-max: @screen-lg-ht-min - 1;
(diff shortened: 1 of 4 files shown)


src/stylesheets/variables.less
```diff
@@ -49,9 +49,13 @@
...
 @font-size-base: 14px;
 
-@screen-lg-ht: 835px;
-@screen-md-ht: 648px;
-@screen-sm-ht: 488px;
+@screen-lg-ht-min: 835px;
+@screen-md-ht-min: 648px;
+@screen-sm-ht-min: 488px;
+
+@screen-md-ht-min-max: @screen-lg-ht-min - 1;
```
mspensieri: `min-max`?

chloepouprom:
Whoa, nice catch

chloepouprom:
Once the `min-max` variable names are fixed, 👍

src/stylesheets/variables.less
lemieux: lol bad find and replace right there

mspensieri:
:+

PR #311, opened 2016-07-06
Add country selector

jugarrit:
@lemieux 
@dannytranlx 
@Mario54 
@alavers


dannytranlx:
Diff is huge, can you point us to files to review?

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -13,13 +14,10 @@ export class TwilioChannelContent extends Component {
...
-        if (isMobile.any) {
-            return <a href={ `sms:${phoneNumber}` }
-                      style={ styleOverride }>
-                       { phoneNumber }
-                   </a>;
-
-        }
-        return <span className='channel-content-value'>{ phoneNumber }</span>;
+        return <TelephoneInput ref='telInput'
+                               defaultCountry='ca'
```
dannytranlx: Should this be configurable? Or have some automated location detection?

lemieux: yeah.. this can't be `ca`... pretty sure we could do something with the user language client-side

src/js/components/telephone-input.jsx
```diff
@@ -0,0 +1,13 @@
+import ReactTelInput from 'react-telephone-input';
+
+export class TelephoneInput extends ReactTelInput {
+    getFlagStyle = () => {
```
lemieux: why not 
```
getFlagStyle() {
    //...
}
```
instead? You don't seem to need to preserve `this` in there.

jugarrit: Good call. I just automatically use fat arrow sometimes.

jugarrit: That's actually not working for some reason @lemieux

lemieux: maybe it has to do with the fact that the original class is using React.createClass and not a ES6 class... that's fine then, use fat arrow in this case

lemieux:
works for me :+1:

src/js/components/channels/twilio-channel-content.jsx
dannytranlx: This diff disappear, but it's still valid? Create a card maybe?

alavers:
Could you open a feature PR without the flags in the diff? Also are we sure we need to check in so many flag svgs? they're going to be on the CDN anyway right?

Also 2, shouldn't there be a single image file with all flags, and not hundreds of separate files?

lemieux:
Not sure why you closed that PR, but I guess @jugarrit didn't add any instructions. You can review that commit https://github.com/smooch/smooch-js/pull/311/commits/3696cae76f0dc5d4ddcccda77aec7a5a2b28a19c which represent pretty much all the changes in this PR. 

Not sure about the single image vs multiple, but they need to be checked out. Especially if they go on the CDN. The build process is taking care of that.

PR #312, opened 2016-07-06
Improve test suite

lemieux:
This is pretty much bringing back tests and adding more that covers mostly what we added with user linking. I also spent some time going over files to adjust styling and adding prop types where needed.

Since we now use the context to pass down some information down the component tree, we wrote a small util that will wrap the component to test in a HOC (higher order component) that will pass the necessary context. Most of the changes are pretty much just about converting existing test cases to use that helper.

I also added a handy check that will make the tests fail if some errors or warning are thrown in the console (like for React proptypes or unhandled promise rejections). That should help us catch error where the tests were ignoring them.

@dannytranlx @Mario54 @chloepouprom @mspensieri @alavers @jugarrit


.eslintrc
```diff
@@ -26,7 +26,9 @@
         "no-constant-condition": 0,
         "no-redeclare": 0,
         "prefer-const": 2,
-        "arrow-parens": [2, "always"]
+        "arrow-parens": [2, "always"],
+        "mocha/no-exclusive-tests": "error",
```
lemieux: This will fail the tests if we forget a `.only` in the tests.

chloepouprom:
Why are some test files in `.js` format while others are in `.jsx`?

test/specs/components/header.spec.js
```diff
@@ -102,12 +96,7 @@ describe('Header', () => {
                 }
             });
             mockedStore = mockAppStore(sandbox, {});
-            parentComponent = TestUtils.renderIntoDocument(<ParentComponentWithContext context={ context }
-                                                                                       store={ mockedStore }
-                                                                                       withRef={ true }>
-                                                               <HeaderComponent {...props} />
-                                                           </ParentComponentWithContext>);
-            header = parentComponent.refs.childElement;
+            header = header = wrapComponentWithContext(HeaderComponent, props, {...context, store: mockedStore});
```
chloepouprom: header = header =

test/utils/react.jsx
```diff
@@ -42,4 +43,15 @@ export function getContext(context = {}) {
         ...defaultContext,
         ...context
     };
-}
\ No newline at end of file
+}
+
+export function wrapComponentWithContext(Component, props, context = getContext()) {
```
chloepouprom: This made the tests so much easier to read. 😁 😁

lemieux:
> Why are some test files in .js format while others are in .jsx?
Those in .jsx have JSX markup in them, those in .js don't. Most of the spec files became JSX-free once they were converted to use the util.

test/specs/components/header.spec.js
lemieux: just wanted to make sure it's assigned :grin:

mspensieri:
👍

chloepouprom:
👍

PR #314, opened 2016-07-11
Use badge color and contrast detection

lemieux:
This pretty much allow the widget to use the new badgeColor field to customize the header. It also brings in constrast detection for better text color where any color customization is used. It means that if the original design was using blue background and white text, but you decided to use a yellow background, the text would switch to black.

Instead of going component per component to handle that customization, I decided to just add classes at the top of the widget and each component style who react based on that. It's much simpler and requires way less changes now and in the future.

Oh, yeah... I also tweak `npm run start-dev` to use nodemon and restart the server whenever an html file changes or a config file changes.

@chloepouprom @dannytranlx @alavers @mspensieri @Mario54 @jugarrit

### DIFF
--- a/src/stylesheets/conversation.less
+++ b/src/stylesheets/conversation.less
@@ -310,3 +310,27 @@
 
     }
 }
+
+.sk-accent-color-light {
+    #sk-conversation {
+        .sk-right-row {
+            .sk-msg, .sk-msg-image {
+                color: @sk-darker-grey;
+
+                a.link, a.link:visited {
+                    color: @sk-darker-grey;
+                }
+            }
+        }
+    }
+}
+
+.sk-link-color-light {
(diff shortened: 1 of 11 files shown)


src/stylesheets/conversation.less
```diff
@@ -310,3 +310,27 @@
 
     }
 }
+
+.sk-accent-color-light {
```
chloepouprom: So `sk-accent-color-dark`, `sk-link-color-dark`, and `sk-branding-color-light` will just follow the default behavior?

lemieux: yeah basically. currently they have no use but they'll be there for future stuff.

chloepouprom:
👍

mspensieri:
👍

PR #315, opened 2016-07-11
Add flag selector to twilio channel

jugarrit:
@alavers 
@lemieux 
@dannytranlx 
@mspensieri 
@liyefei737


jugarrit:
Added a `png` with the full list of flags.

Added a new vendored library, `react-telephone-input` directly into libs folder for the following reasons:

- Much easier to customize. The library had a lot of options but I wanted to add better `onInputValid` functionality and strip out code we weren't using to lessen our footprint.
- It made use of lots of lodash functions that are built in to ES6, these have been replaced
- etc.

Updated the SMS description.

Created new mixins to automatically set the width of a div based on the current size of the window.

This will allow us to have content correctly match the available width without having to copy/paste a bunch.

lemieux:
:+1:

src/stylesheets/flags.less
```diff
@@ -0,0 +1,796 @@
+@flag-width: 32px;
+@flag-height: 24px;
+
+// @flag-width: 80px;
+// @flag-height: 60px;
```
dannytranlx: Comments?

dannytranlx:
👍

src/stylesheets/flags.less
jugarrit: whupz

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,25 +1,34 @@
...
+            numberValid: false
+        });
+    }
 
-        }
-        return <span className='channel-content-value'>{ phoneNumber }</span>;
+    render() {
+        return <div>
+                   <ReactTelephoneInput ref='telInput'
+                                        defaultCountry='ca'
```
alavers: should be us perhaps?

alavers:
👍

src/js/components/channels/twilio-channel-content.jsx
dannytranlx: we talked about having it configurable in some previous PRs, don't know what was the outcome of that discussion

PR #316, opened 2016-07-11
Floating widget

lemieux:
As per new design, the widget shall now float! Here's some screenshots :

Large widget : 
image

Small widget : 
image

Mobile (still taking 100% space):
image

Embedded (still taking 100% space of container) :
image

@dannytranlx @mspensieri @alavers @jugarrit @chloepouprom

### DIFF
--- a/src/stylesheets/main.less
+++ b/src/stylesheets/main.less
@@ -31,6 +31,24 @@
     @import "visibility.less";
     @import (css) "~react-spinkit/css/three-bounce.css";
 }
+
+.widget-size() {
+    width: @widget-width-md;
+    height: @widget-height-md;
+    @media (min-width: @screen-lg-min) and (min-height: @screen-md-ht-min) {
+        width: @widget-width-lg;
+        height: @widget-height-lg;
+    }
+    @media (min-width: @screen-sm-min) and (max-height: @screen-sm-ht-min) {
+        width: @widget-width-sm;
+        height: @widget-height-sm;
+    }
+    @media (max-width: @screen-sm-min) {
+        width: @widget-width-xs;
+        height: @widget-height-xs;
+    }
+}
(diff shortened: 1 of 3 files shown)


Mario54:
What's up with the "send" button in the first screenshot?

mspensieri:
👍

src/stylesheets/main.less
```diff
@@ -31,13 +31,31 @@
     @import "visibility.less";
     @import (css) "~react-spinkit/css/three-bounce.css";
 }
+
+.widget-size() {
```
chloepouprom: 👌 🙌

chloepouprom:
So, for tab version, are just gonna add a new class (something like `sk-tab`)?

lemieux:
> What's up with the "send" button in the first screenshot?

eh, not sure, I'll double-check.

> So, for tab version, are just gonna add a new class (something like sk-tab)?

you're so impatient, wait for my next PR ;) I'm doing all one this piece by piece and in this PR the badge mode doesn't exist yet.

src/stylesheets/main.less
```diff
@@ -7,7 +7,7 @@
     font-weight: 400;
     font-size: 13px;
     line-height: 1.4;
-    border-radius: 10px 10px 0 0;
+    border-radius: 10px 10px 10px 10px;
```
dannytranlx: Would writing `10px` be applied to all 4 corners?

Let save all the kBs 😂

lemieux: yeah you're right

dannytranlx:
👍

dannytranlx:
@chloepouprom On the first screenshot, didn't we fix that with some React wizardery? There's an huge padding on the right.

PR #317, opened 2016-07-13
Badge mode

lemieux:
This bring badge mode support for the widget.

Since the animation is using scaling, the input was getting 0px as its width on initial render. I fixed that by using flexboxes instead (which I tested on Chrome, iOS Safari, Chrome Android, and IE10 so far). That also fixes the behavior we saw in embedded mode and removes the need of doing it in JS.

@chloepouprom @dannytranlx @jugarrit @mspensieri @alavers

### DIFF
--- a/src/js/components/badge.jsx
+++ b/src/js/components/badge.jsx
@@ -1,11 +1,12 @@
 import React, { Component, PropTypes } from 'react';
 
+import { openWidget } from '../services/app-service';
+
 class DefaultIcon extends Component {
     render() {
-        const {badgeColorDark} = this.props;
+        const {brandColorDark} = this.props;
         const style = {
-            filter: `drop-shadow( -5px -5px 5px #${badgeColorDark ? 'fff': '000'} )`,
-            fill: badgeColorDark ? '#fff' : '#000'
+            fill: brandColorDark ? '#fff' : '#000'
         };
 
         return <svg version='1.0'
@@ -26,10 +27,22 @@ export class Badge extends Component {
         settings: PropTypes.object.isRequired
     };
 
+    onClick = (e) => {
+        e.preventDefault();
+        openWidget();
+    };
+
     render() {
-        const {settings: {badgeColor, badgeColorDark, badgeIconUrl}} = this.context;
-        return <div className='sk-badge'>
-                   <DefaultIcon badgeColorDark={ badgeColorDark } />
+        const {settings: {brandColor, brandColorDark, badgeIconUrl}} = this.context;
+
+        const style = {
+            backgroundColor: `#${brandColor}`
+        };
+
+        return <div className='sk-badge'
+                    style={ style }
+                    onClick={ this.onClick }>
+                   <DefaultIcon brandColorDark={ brandColorDark } />
                </div>;
     }
 }
--- a/src/js/components/badge.jsx
+++ b/src/js/components/badge.jsx
@@ -43,6 +43,14 @@ export class Badge extends Component {
         settings: PropTypes.object.isRequired
     };
 
+    static propTypes = {
+        shown: PropTypes.bool
+    };
+
+    static defaultProps = {
+        shown: true
+    };
+
     onClick = (e) => {
         e.preventDefault();
         openWidget();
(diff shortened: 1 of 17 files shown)


src/stylesheets/animations.less
```diff
@@ -1,209 +1,207 @@
...
     }
-}
 
-.sk-close {
-    bottom: @widget-close-bottom-md;
+    &.sk-close {
+        bottom: @widget-close-bottom-md;
 
-    -webkit-animation: sk-close-frames-md .4s cubic-bezier(.62, .28, .23, .99);
             animation: sk-close-frames-md .4s cubic-bezier(.62, .28, .23, .99);
```
chloepouprom: Are all the `sk-close-frames` related to the tab widget? If so, I would rename to `sk-close-tab-frames` to be consistent with the `sk-close-badge-frames`

lemieux: good point

lemieux:
Small comment : I removed all the -webkit prefix from animations because it seems they are useless now.

src/js/components/connect-notification.jsx
```diff
@@ -66,9 +66,9 @@ export class ConnectNotificationComponent extends Component {
                     const separator = index !== array.length - 1 ? ',' : '';
 
                     return <div style={ linkStyle }
-                                className='channel-details'>
-                               <a key={ channel.type }
-                                  style={ linkStyle }
+                                className='channel-details'
+                                key={channel.type}>
```
chloepouprom: Why did you move the key from `<a>` to `<div>`?

lemieux: because the key needs to be at the top level element of an item in a list.

lemieux: it was throwing warning about it

chloepouprom:
👍

src/js/components/badge.jsx
```diff
@@ -0,0 +1,87 @@
+import React, { Component, PropTypes } from 'react';
+
+import { openWidget } from '../services/app-service';
+import { SK_DARK_CONTRAST } from '../constants/styles';
+
+export class DefaultIcon extends Component {
```
Mario54: nit: rename `DefaultBadgeIcon`?

lemieux: sure, but your're already in the "badge" module /shrug

src/js/components/widget.jsx
```diff
@@ -105,39 +109,46 @@ export class WidgetComponent extends Component {
...
-                    { channelsComponent }
-                    <Conversation />
-                    { footer }
-                </div>
-            </div>
-            );
+        let badge;
+
+        if (displayStyle === DISPLAY_STYLE.BADGE && !appState.embedded) {
+            badge = <Badge shown={ !appState.widgetOpened } />;
```
Mario54: Why is it passing a `shown` prop instead of setting `badge` to `null` if widgetOpened is true?

src/js/constants/styles.js
```diff
@@ -3,5 +3,9 @@ export const ANIMATION_TIMINGS = {
 };
 
 export const SK_PURPLE = '#92278f';
+export const SK_DARK_CONTRAST = '#55555d';
 
-export const SK_DARK_CONSTRAST = '#55555d';
+export const DISPLAY_STYLE = {
+    BADGE: 'badge',
```
Mario54: I wonder if we should use `Symbol()` instead of 'badge' or 'tab' here?

lemieux: why would we? it matches string values returned by the server

src/js/components/widget.jsx
lemieux: animation

src/js/constants/styles.js
Mario54: oh I thought it was just constants for the widget

src/js/components/badge.jsx
Mario54: Yeah but it's exported and you said that export names should stand on their own :)

Mario54:
👍

PR #319, opened 2016-07-13
Reorder operations when destroying widget

lemieux:
The widget was throwing errors on destroy because the Introduction component is recomputing  its height on update (`store.dispatch(reset());` was triggering an unnecessary update), but it's debounced so that computation was happening after the widget was unmounted. By rearranging the order of action, we ensure the the widget is properly unmounted and doesn't listen to the store before resetting anything.

@dannytranlx @jugarrit @mspensieri @Mario54

### DIFF
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -294,13 +294,16 @@ export class Smooch {
             console.warn('Smooch.destroy was called before Smooch.init was called properly.');
         }
 
-        const {embedded} = store.getState().appState;
-        disconnectFaye();
-        store.dispatch(reset());
+        stopMonitoringBrowserState();
+        
         if (process.env.NODE_ENV !== 'test' && this._container) {
             unmountComponentAtNode(this._container);
         }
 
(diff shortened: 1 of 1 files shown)


src/js/smooch.jsx
```diff
@@ -294,13 +294,16 @@ export class Smooch {
...
+        stopMonitoringBrowserState();
+        
         if (process.env.NODE_ENV !== 'test' && this._container) {
             unmountComponentAtNode(this._container);
         }
 
+        disconnectFaye();
+        store.dispatch(reset());
+
+        const {embedded} = store.getState().appState;
```
jugarrit: I guess this state will still be correct even though you're resetting right before?

jugarrit:
👍

src/js/smooch.jsx
lemieux: ehh good catch, it should go before the reset.

chloepouprom:
👍

PR #320, opened 2016-07-13
Contrast building color with background color

lemieux:
It's pretty much a must fix for existing user since they'll have brandColor = #fff and it would end up white on white for the business icon. This now uses the same logic as other parts of the UI to contrast the content of the icon.

@mspensieri @dannytranlx @Mario54 @jugarrit @alavers @chloepouprom

### DIFF
--- a/src/js/components/default-app-icon.jsx
+++ b/src/js/components/default-app-icon.jsx
@@ -2,10 +2,6 @@ import React, { Component, PropTypes } from 'react';
 import { SK_DARK_CONSTRAST } from '../constants/styles';
 
 export class DefaultAppIcon extends Component {
-    static propTypes = {
-        color: PropTypes.string
-    };
-
     static contextTypes = {
         settings: PropTypes.object.isRequired
     };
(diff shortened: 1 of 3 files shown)


jugarrit:
👍

src/js/components/default-app-icon.jsx
```diff
@@ -1,68 +1,73 @@
...
+import React, { Component, PropTypes } from 'react';
+import { SK_DARK_CONSTRAST } from '../constants/styles';
 
 export class DefaultAppIcon extends Component {
-    static defaultProps = {
-        style: {
-            fill: SK_PURPLE
-        }
+    static propTypes = {
+        color: PropTypes.string
```
chloepouprom: Do we still need this? Since the color is in the context

Mario54:
👍

src/js/components/default-app-icon.jsx
lemieux: eh good catch, it should go away

PR #323, opened 2016-07-18
Apply contrast detection to settings header

lemieux:
Since the normal header and the settings header are not using the same id, the contrast logic wasn't applied to the settings and channels pages.

@chloepouprom @Mario54 @mspensieri @alavers @dannytranlx


mspensieri:
👍

alavers:
👍

PR #325, opened 2016-07-25
Add UI for SMS linking

chloepouprom:
@jugarrit @liyefei737 @lemieux @dannytranlx 

screen shot 2016-07-25 at 11 59 23 am

screen shot 2016-07-25 at 11 59 29 am

screen shot 2016-07-25 at 11 59 38 am

And on mobile (opens up SMS app):

To-do:
- Improved number validation
- API calls

### DIFF
--- a/src/js/components/channels/twilio-channel-content.jsx
+++ b/src/js/components/channels/twilio-channel-content.jsx
@@ -94,32 +94,26 @@ export class TwilioChannelContentComponent extends Component {
                                  </div>;
 
         const sendTextUrl = `sms://${phoneNumber}`;
+        const linkedComponentButton = isMobile.phone ? <button className='btn btn-sk-primary twilio-linking'
+                                                               onClick={ this.onStartTexting }>
+                                                           Start Texting
+                                                       </button> :
+            <button className='btn btn-sk-primary twilio-linking'
+                    onClick={ this.onSendText }>
+                Send me a text
+            </button>;
 
-        const mobileLinkedComponent = <div>
-                                          <div className='twilio-linking unconfirmed-state'>
-                                              <i className='fa fa-phone'
-                                                 style={ iconStyle }></i>
-                                              <span className='phone-number'>{ number }</span>
-                                              <a onClick={ this.onChangeNumber }>Change my number</a>
-                                          </div>
-                                          <a href={ sendTextUrl }>
-                                              <button className='btn btn-sk-primary twilio-linking'
-                                                      onClick={ this.onStartTexting }> Start Texting </button>
-                                          </a>
-                                      </div>;
-        const desktopLinkedComponent = <div>
-                                           <div className='twilio-linking linked-state'>
-                                               <i className='fa fa-phone'
-                                                  style={ iconStyle }></i>
-                                               <span className='phone-number'>{ number }</span>
-                                               <a onClick={ this.onChangeNumber }>Change my number</a>
-                                           </div>
-                                           <a href={ sendTextUrl }>
-                                               <button className='btn btn-sk-primary twilio-linking'
-                                                       onClick={ this.onSendText }> Send me a text </button>
-                                           </a>
-                                       </div>;
-        const linkedComponent = isMobile.phone ? mobileLinkedComponent : desktopLinkedComponent;
+        const linkedComponent = <div>
+                                    <div className='twilio-linking linked-state'>
+                                        <i className='fa fa-phone'
+                                           style={ iconStyle }></i>
+                                        <span className='phone-number'>{ number }</span>
+                                        <a onClick={ this.onChangeNumber }>Change my number</a>
+                                    </div>
+                                    <a href={ sendTextUrl }>
+                                        { linkedComponentButton }
+                                    </a>
+                                </div>;
         if (linkState === 'pending') {
             return pendingComponent;
         } else if (linkState === 'linked') {
--- a/src/js/components/channels/twilio-channel-content.jsx
+++ b/src/js/components/channels/twilio-channel-content.jsx
@@ -59,17 +59,18 @@ export class TwilioChannelContentComponent extends Component {
     }
 
     render() {
+        const {number, numberValid, phoneNumber, linkState, settings: {linkColor}} = this.props;
         let iconStyle = {};
-        if (this.props.settings.linkColor) {
+        if (linkColor) {
             iconStyle = {
-                color: `#${this.props.settings.linkColor}`
+                color: `#${linkColor}`
             };
         }
 
-        const linkButton = this.props.numberValid ? <button className='btn btn-sk-primary'
-                                                            onClick={ this.linkTwilioNumber }>
-                                                        Continue
-                                                    </button> : '';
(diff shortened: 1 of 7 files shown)


src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,134 @@
...
+                color: `#${this.props.settings.linkColor}`
+            };
+        }
+
+        const linkButton = this.props.numberValid ? <button className='btn btn-sk-primary'
+                                                            onClick={ this.linkTwilioNumber }>
+                                                        Continue
+                                                    </button> : '';
+        const unlinkedComponent = <div>
+                                      <ReactTelephoneInput ref='telInput'
```
lemieux: string refs are legacy since React 15 (https://facebook.github.io/react/docs/more-about-refs.html). The new way to use them is to use callbacks like this `<ReactTelephoneInput ref={(c) => this._telInput = c} /> ` and then use `this._telInput` directly.

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,134 @@
...
+                                                           defaultCountry='ca'
+                                                           onChange={ this.handleInputChange }
+                                                           onValid={ this.onNumberValid }
+                                                           onInvalid={ this.onNumberInvalid }
+                                                           preferredCountries={ ['ca', 'us'] }
+                                                           onBlur={ this.handleInputBlur } />
+                                      { linkButton }
+                                  </div>;
+        const pendingComponent = <div>
```
lemieux: there's an unnecessary wrapping div around this.

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,134 @@
...
+                                                  style={ iconStyle }></i>
+                                               <span className='phone-number'>{ this.props.number }</span>
+                                               <a onClick={ this.onChangeNumber }>Change my number</a>
+                                           </div>
+                                           <a href={ sendTextUrl }>
+                                               <button className='btn btn-sk-primary twilio-linking'
+                                                       onClick={ this.onSendText }> Send me a text </button>
+                                           </a>
+                                       </div>;
+        const linkedComponent = isMobile.any ? mobileLinkedComponent : desktopLinkedComponent;
```
lemieux: `isMobile.any` will also match tablets

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,134 @@
...
-                   <ReactTelephoneInput ref='telInput'
-                                        defaultCountry='ca'
-                                        onChange={ this.handleInputChange }
-                                        onValid={ this.onNumberValid }
-                                        onInvalid={ this.onNumberInvalid }
-                                        preferredCountries={ ['ca', 'us'] }
-                                        onBlur={ this.handleInputBlur } />
-               </div>;
+        let iconStyle = {};
+        if (this.props.settings.linkColor) {
```
lemieux: I see a lot of `this.props.something` all over this method, can we get them from the start with `const { ...} = this.props` and avoid repetition?

src/stylesheets/channels.less
```diff
@@ -73,3 +73,34 @@
...
+            font-size: smaller;
+            text-decoration: underline;
+            float: right;
+            padding-right: 3px;
+        }
+
+        &.btn {
+            margin-top: 25px;
+            text-align: center;
+            border: 0px;
```
lemieux: We have a button in email settings too, they should match imo

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,133 @@
...
-export class TwilioChannelContent extends Component {
-    state = {
-        numberValid: false
+import { connect } from 'react-redux';
+import isMobile from 'ismobilejs';
+
+export class TwilioChannelContentComponent extends Component {
+
+    linkTwilioNumber = () => {
+        const {state: {formattedNumber}} = this._telInput;
```
lemieux: I'm pretty sure it's against React encapsulation logic to acces a component state directly. I think you should have an `onChange` callback instead and track the value from there. https://stackoverflow.com/questions/27849319/react-is-accessing-the-state-of-children-an-anti-pattern https://facebook.github.io/react/tips/communicate-between-components.html

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,133 @@
...
+                                     <i className='fa fa-phone'
+                                        style={ iconStyle }></i>
+                                     <span className='phone-number'>{ number } - Pending</span>
+                                     <a onClick={ this.onRetry }>Retry</a>
+                                 </div>;
+
+        const sendTextUrl = `sms://${phoneNumber}`;
+
+        const mobileLinkedComponent = <div>
+                                          <div className='twilio-linking unconfirmed-state'>
```
jugarrit: Why is this `unconfirmed-state` if  it's linked?

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,133 @@
...
+        const {state: {formattedNumber}} = this._telInput;
+        updateTwilioAttributes({
+            number: formattedNumber,
+            linkState: 'pending'
+        });
     }
 
+    onRetry = () => {
+        updateTwilioAttributes({
+            linkState: 'linked'
```
jugarrit: Shouldn't this set the state to 'unlinked'?

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,34 +1,138 @@
...
+                                      </div>;
+        const desktopLinkedComponent = <div>
+                                           <div className='twilio-linking linked-state'>
+                                               <i className='fa fa-phone'
+                                                  style={ iconStyle }></i>
+                                               <span className='phone-number'>{ number }</span>
+                                               <a onClick={ this.onChangeNumber }>Change my number</a>
+                                           </div>
+                                           <a href={ sendTextUrl }>
+                                               <button className='btn btn-sk-primary twilio-linking'
```
lemieux: I'll guess we'll wait for me to merge the new settings, but we need to remember that these buttons will need to use the link color and contrast detection

src/js/components/channels/twilio-channel-content.jsx
chloepouprom: oops good catch

src/js/components/channels/twilio-channel-content.jsx
chloepouprom: yep! I only set it to `linked` so that I could test the UI. will fix

jugarrit:
We need to determine what the correct state is at launch as well from what is returned in the `init` call. I assume this will be part of the next PR?

chloepouprom:
@jugarrit yep! I'll create a separate PR for that.

jugarrit:
Okiedokes. lgtm then

lemieux:
:+1:

PR #326, opened 2016-07-25
Auto deploy integration to staging website

lemieux:
This will auto-deploy the integration branch whenever it gets a new commit to a staging website (which I'll give the link in slack). The only difference between the staging build and the production build is that the staging build is inlining all its resources to avoid having to mess around with cdn and stuff.

@alavers @dannytranlx @mspensieri @jmtremblay

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -27,7 +27,7 @@ deployment:
       - grunt deploy
 
   staging:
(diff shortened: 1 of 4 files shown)


mspensieri:
👍

PR #328, opened 2016-07-26
Update Twilio integration state with user information

chloepouprom:
The state is now updated with the correct Twilio-related information. That way, if the user already has SMS linking enabled, the right view will be displayed.

I also added a few frontend tests!

@jugarrit @liyefei737 @lemieux @Mario54 @mspensieri

### DIFF
--- a/src/js/services/integrations-service.js
+++ b/src/js/services/integrations-service.js
@@ -35,13 +35,19 @@ export function resetTwilioAttributes() {
 }
 
 export function fetchTwilioAttributes() {
-    const {user: {clients}} = store.getState();
+    const {user: {clients, pendingClients}} = store.getState();
     const client = clients.find((client) => client.platform === 'twilio');
+    const pendingClient = pendingClients.find((client) => client.platform === 'twilio');
 
     if (client) {
         updateTwilioAttributes({
             linkState: 'linked',
             number: client.info.phoneNumber
+        });
+    } else if (pendingClient) {
+        updateTwilioAttributes({
+            linkState: 'pending',
+            number: pendingClient.info.phoneNumber
         });
     }
 }
--- a/src/js/services/integrations-service.js
+++ b/src/js/services/integrations-service.js
@@ -42,12 +42,12 @@ export function fetchTwilioAttributes() {
     if (client) {
         updateTwilioAttributes({
             linkState: 'linked',
-            number: client.info.phoneNumber
+            number: client.displayName
         });
     } else if (pendingClient) {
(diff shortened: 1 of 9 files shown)


src/js/services/integrations-service.js
```diff
@@ -33,3 +33,15 @@ export function updateTwilioAttributes(attr) {
 export function resetTwilioAttributes() {
     store.dispatch(resetTwilioIntegrationState());
 }
+
+export function fetchTwilioAttributes() {
```
jugarrit: May want to add the case for pending in here.

`const pendingClient = pendingClients.find((client) => client.platform === 'twilio');`

src/js/utils/user.js
```diff
@@ -7,7 +7,11 @@ export function isChannelLinked(clients, channelType) {
 
 export function getDisplayName(clients, channelType) {
     const client = clients.find((client) => client.platform === channelType);
-    return client && client.displayName;
+    if (channelType === 'twilio') {
```
jugarrit: Hmmm maybe the backend should just be returning the correct `displayName` for Twilio?

test/specs/components/channels/twilio-channel-component.spec.jsx
```diff
@@ -0,0 +1,93 @@
...
+            numberValid: false,
+            settings: {}
+        };
+
+        it('should render unlinked component', () => {
+            component = TestUtils.renderIntoDocument(<TwilioChannelContentComponent {...unlinkedProps}/>);
+            TestUtils.scryRenderedDOMComponentsWithClass(component, 'mockedTelephoneInput').length.should.eq(1);
+        });
+
+        [true, false].forEach((numberValid) => {
```
jugarrit: Neat.

src/js/utils/user.js
chloepouprom: Yeah, that makes more sense! I'll add a task for that

src/js/services/integrations-service.js
```diff
@@ -35,13 +35,19 @@ export function resetTwilioAttributes() {
...
 
     if (client) {
         updateTwilioAttributes({
             linkState: 'linked',
             number: client.info.phoneNumber
         });
+    } else if (pendingClient) {
+        updateTwilioAttributes({
+            linkState: 'pending',
+            number: pendingClient.info.phoneNumber
```
jugarrit: There won't actually be any info on the `pendingClient` here. We'll have to look in the app `integrations` list.

mspensieri:
👍

PR #332, opened 2016-07-29
Fix disappearing widget

chloepouprom:
Widget was not there on screens of really small height when it was in `init` state

@lemieux


lemieux:
:+1:

jugarrit:
👍

jugarrit:
We'll need to resolve conflicts when the sms-linking comes in... Whoo

PR #333, opened 2016-08-01
Make call to linking API

chloepouprom:
@jugarrit @lemieux @alavers @mspensieri 

This adds calls to API endpoints from https://github.com/smooch/smooch-core-js/pull/37

TODO:
- Improved error handling
- Improved number validation

### DIFF
--- a/src/js/components/channels/twilio-channel-content.jsx
+++ b/src/js/components/channels/twilio-channel-content.jsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { Component, PropTypes } from 'react';
 import { updateTwilioAttributes, resetTwilioAttributes, linkTwilioChannel, deleteTwilioChannel } from '../../services/integrations-service';
 
 import { ReactTelephoneInput } from '../../lib/react-telephone-input';
@@ -7,6 +7,10 @@ import { connect } from 'react-redux';
 import isMobile from 'ismobilejs';
 
 export class TwilioChannelContentComponent extends Component {
+
+    static contextTypes = {
+        settings: PropTypes.object
+    };
 
     linkTwilioNumber = () => {
         linkTwilioChannel(this.props.userId, {
@@ -62,7 +66,8 @@ export class TwilioChannelContentComponent extends Component {
     }
 
     render() {
-        const {appUserNumber, appUserNumberValid, phoneNumber, linkState, settings: {linkColor}} = this.props;
+        const {appUserNumber, appUserNumberValid, phoneNumber, linkState} = this.props;
+        const {settings: {linkColor}} = this.context;
         let iconStyle = {};
         if (linkColor) {
             iconStyle = {
@@ -125,7 +130,6 @@ export class TwilioChannelContentComponent extends Component {
 export const TwilioChannelContent = connect((state) => {
     return {
         ...state.integrations.twilio,
-        settings: state.app.settings.web,
         userId: state.user._id
     };
 })(TwilioChannelContentComponent);
--- a/src/js/components/channels/twilio-channel-content.jsx
+++ b/src/js/components/channels/twilio-channel-content.jsx
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { updateTwilioAttributes, resetTwilioAttributes } from '../../services/integrations-service';
+import { updateTwilioAttributes, resetTwilioAttributes, linkChannel, deleteChannel } from '../../services/integrations-service';
 
 import { ReactTelephoneInput } from '../../lib/react-telephone-input';
 
@@ -9,15 +9,15 @@ import isMobile from 'ismobilejs';
 export class TwilioChannelContentComponent extends Component {
 
     linkTwilioNumber = () => {
-        updateTwilioAttributes({
-            appUserNumber: this.props.appUserNumber,
-            linkState: 'pending'
+        linkChannel(this.props.userId, {
+            type: 'twilio', 
+            phoneNumber: this.props.appUserNumber.replace(/[()\-\s]/g, '')
         });
     }
 
-    onRetry = () => {
-        updateTwilioAttributes({
-            linkState: 'unlinked'
+    deleteChannel = () => {
+        deleteChannel(this.props.userId, {
+            type: 'twilio'
         });
     }
 
@@ -90,7 +90,7 @@ export class TwilioChannelContentComponent extends Component {
                                      <i className='fa fa-phone'
                                         style={ iconStyle }></i>
                                      <span className='phone-number'>{ appUserNumber } - Pending</span>
-                                     <a onClick={ this.onRetry }>Retry</a>
+                                     <a onClick={ this.deleteChannel }>Retry</a>
                                  </div>;
 
         const sendTextUrl = `sms://${phoneNumber}`;
@@ -108,7 +108,7 @@ export class TwilioChannelContentComponent extends Component {
                                         <i className='fa fa-phone'
                                            style={ iconStyle }></i>
                                         <span className='phone-number'>{ appUserNumber }</span>
-                                        <a onClick={ this.onChangeNumber }>Change my number</a>
+                                        <a onClick={ this.deleteChannel }>Change my number</a>
                                     </div>
                                     <a href={ sendTextUrl }>
                                         { linkedComponentButton }
@@ -127,6 +127,7 @@ export class TwilioChannelContentComponent extends Component {
 export const TwilioChannelContent = connect((state) => {
     return {
         ...state.integrations.twilio,
-        settings: state.app.settings.web
+        settings: state.app.settings.web,
+        userId: state.user._id
     };
 })(TwilioChannelContentComponent);
--- a/src/js/components/channels/twilio-channel-content.jsx
+++ b/src/js/components/channels/twilio-channel-content.jsx
@@ -1,5 +1,5 @@
 import React, { Component, PropTypes } from 'react';
-import { updateTwilioAttributes, resetTwilioAttributes, linkTwilioChannel, deleteTwilioChannel } from '../../services/integrations-service';
+import { updateTwilioAttributes, resetTwilioAttributes, linkTwilioChannel, unlinkTwilioChannel } from '../../services/integrations-service';
 
 import { ReactTelephoneInput } from '../../lib/react-telephone-input';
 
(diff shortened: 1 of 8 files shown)


src/js/services/integrations-service.js
```diff
@@ -37,17 +38,55 @@ export function resetTwilioAttributes() {
...
         });
     }
 }
+
+export function linkTwilioChannel(userId, data) {
+    return core().appUsers.link.linkChannel(userId, data)
+        .then((appUser) => {
+            immediateUpdate(appUser).then(() => {
+                updateTwilioAttributes({
+                    linkState: 'pending'
```
jugarrit: I might update the `appUserNumber` here to match what's returned (say we validate the number but store a slightly different format)

I'm not sure if this will be applicable though, just worth mentioning.

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,55 @@ export function resetTwilioAttributes() {
...
+                });
+            });
+        })
+        .catch(() => {
```
jugarrit: Check the error code here. 
If it's `4xx`, I doubt retrying would help. Maybe a different message like: `We were unable to communicate with this number, please enter a different one.` or something. Copy to be reviewed.

jugarrit:
Your test coverage is always top notch.
Good work 👍

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -127,6 +125,7 @@ export class TwilioChannelContentComponent extends Component {
 export const TwilioChannelContent = connect((state) => {
     return {
         ...state.integrations.twilio,
-        settings: state.app.settings.web
+        settings: state.app.settings.web,
```
lemieux: settings are part of the context, you don't need to add them in `connect`.

lemieux: I guess it's not your change, but /shrug

chloepouprom: heh yeah. I added this in one of my previous PRs. I'll fix this! 😁

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,55 @@ export function resetTwilioAttributes() {
...
-            number: pendingClient.displayName
+            appUserNumber: pendingClient.displayName
         });
     }
 }
+
+export function linkTwilioChannel(userId, data) {
+    return core().appUsers.link.linkChannel(userId, data)
+        .then((appUser) => {
+            immediateUpdate(appUser).then(() => {
```
mspensieri: should `return` here, no?

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,55 @@ export function resetTwilioAttributes() {
...
+                hasError: true,
+                errorMessage: 'Something went wrong. Please try again.'
+            });
+        });
+}
+
+export function deleteTwilioChannel(userId) {
+    return core().appUsers.link.deleteTwilioChannel(userId)
+        .then(() => {
+            immediateUpdate({
```
mspensieri: same comment, `return`?

src/js/services/integrations-service.js
lemieux: I would return thise promise so that `linkTwilioChannel` resolves when it's all done. You could even simplify that by doing, this 

```
return core().appUsers.link.linkChannel(userId, data)
    .then(immediateUpdate)
    .then(() => {
        updateTwilioAttributes({
            linkState: 'pending'
        });
    });
```

src/js/services/integrations-service.js
lemieux: this should be returned

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,55 @@ export function resetTwilioAttributes() {
...
+        })
+        .catch(() => {
+            updateTwilioAttributes({
+                hasError: true,
+                errorMessage: 'Something went wrong. Please try again.'
+            });
+        });
+}
+
+export function deleteTwilioChannel(userId) {
```
lemieux: why is this called `delete` when the other action is `link`? Shouldn't this be `unlink`?

lemieux:
I think you will need to change the EDITABLE_PROPERTIES array in user-service, otherwise your user will never be considered dirty and won't be saved. 

Unless I misunderstood what `pendingClients` is supposed to be.

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,55 @@ export function resetTwilioAttributes() {
...
+    return core().appUsers.link.deleteTwilioChannel(userId)
+        .then(() => {
+            immediateUpdate({
+                pendingClients: []
```
lemieux: why is the client setting the pending clients to an empty array? shouldn't the server return the clients status on unlink?

chloepouprom: the server only returns a 200 I think. I re-set the pendingClients to empty. although, I'm not sure if it's possible to have more than one pendingClient... @jugarrit ?

lemieux: If the server returns the appUser on link, I would expect it to return it on unlink too. Also, it's not possible to have more than one *twilio* pending client, but what happens when we have pending email client too?

jugarrit: > If the server returns the appUser on link, I would expect it to return it on unlink too
This isn't the case with most of the `DELETE` APIs we have, is it?

> Also, it's not possible to have more than one twilio pending client, but what happens when we have pending email client too

This is a valid point, we should only remove the `pendingClient` whose platform matches the one we're trying to remove.

lemieux: Then I guess you are only trying to replicate the server side state? You shouldn't be using immediate update then, you should just be updating the store.

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,57 @@ export function resetTwilioAttributes() {
...
-            number: pendingClient.displayName
+            appUserNumber: pendingClient.displayName
         });
     }
 }
+
+export function linkTwilioChannel(userId, data) {
+    return core().appUsers.link.linkChannel(userId, data)
+        .then((appUser) => {
+            return immediateUpdate(appUser);
```
lemieux: why are we updating the user at all? if the server is returning the latest version of the appUser we should just be dispatching it to the store and that's it? There's nothing to be updated, right?

chloepouprom: Nope, nothing should be updated. I removed the calls to immediateUpdate

chloepouprom:
@mspensieri @lemieux any more comments to add on the new changes? 👀

mspensieri:
👍

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,59 @@ export function resetTwilioAttributes() {
 export function fetchTwilioAttributes() {
     const {user: {clients, pendingClients}} = store.getState();
     const client = clients.find((client) => client.platform === 'twilio');
-    const pendingClient = pendingClients.find((client) => client.platform === 'twilio');
+    const pendingClient = pendingClients && pendingClients.find((client) => client.platform === 'twilio');
```
lemieux: is there any case where pendingClients doesn't exist?

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,59 @@ export function resetTwilioAttributes() {
...
-            number: pendingClient.displayName
+            appUserNumber: pendingClient.displayName
         });
     }
 }
+
+export function linkTwilioChannel(userId, data) {
+    return core().appUsers.link.linkChannel(userId, data)
+        .then((appUser) => {
+            store.dispatch(updateUser(...appUser));
```
lemieux: not sure you really need to copy the object here, the reducer should always be doing that

src/js/services/integrations-service.js
```diff
@@ -37,17 +38,59 @@ export function resetTwilioAttributes() {
...
+        })
+        .then(() => {
+            updateTwilioAttributes({
+                linkState: 'pending'
+            });
+        })
+        .catch(() => {
+            updateTwilioAttributes({
+                hasError: true,
```
lemieux: is `hasError` ever cleared?

src/js/services/integrations-service.js
chloepouprom: I got an error at some point... but that's only because I was working with the wrong backend. `pendingClients` should always exist from now on.

src/js/services/user-service.js
```diff
@@ -17,7 +17,9 @@ export const EDITABLE_PROPERTIES = [
     'surname',
     'email',
     'signedUpAt',
-    'properties'
+    'properties',
+    'pendingClients',
```
lemieux: not sure that `clients` or `pendingClients` should be update-able by the client, it's really not something the user should have control on, right? Is the API allowing it anyway?

src/js/services/integrations-service.js
chloepouprom: if I don't, reducer thinks I'm updating the user with key `appUser`

src/js/services/user-service.js
chloepouprom: true. I had added this back then when I was using `immediateUpdate`. will remove

src/js/services/integrations-service.js
lemieux: I don't understand how spreading it makes it different, the problem here is that the server returns and object with `appUser` in it....  the fix here should be to do this 

 ```
export function linkTwilioChannel(userId, data) {
    return core().appUsers.link.linkChannel(userId, data)
        .then(({appUser}) => {
            store.dispatch(updateUser(appUser));
        });
}
```

File: src/js/services/integrations-service.js (hunk 11 above)
Comment (chloepouprom) [inline src/js/services/integrations-service.js]: Covering this in next PR for error handling

File: src/js/services/integrations-service.js [hunk 13]
```diff
@@ -37,17 +38,59 @@ export function resetTwilioAttributes() {
...
+                errorMessage: 'We were unable to communicate with this number. Please enter a different one.'
+            });
+        });
+}
+
+export function unlinkTwilioChannel(userId) {
+    return core().appUsers.link.unlinkChannel(userId, 'twilio')
```
lemieux: I guess I missed that on the smooch-core PR, but why isn't it just `core().appUsers.linkChannel` and `core().appUsers.unlinkChannel` ? It makes a really weird API :(

chloepouprom: 😞

lemieux:
:+1: otherwise

PR #335, opened 2016-08-02
Make #sk-wrapper a flexbox

spasiu:
[*NOTE: have to fix animation before merging. PR'd to get feedback on approach]

@lemieux @dannytranlx @chloepouprom 

Switch from fixed width notification containers to notifications that are sized according to content.

screen shot 2016-08-02 at 12 18 40 pm

screen shot 2016-08-02 at 12 18 57 pm

screen shot 2016-08-02 at 12 19 16 pm

### DIFF
--- a/src/js/components/error-notification.jsx
+++ b/src/js/components/error-notification.jsx
@@ -17,8 +17,7 @@ export class ErrorNotificationComponent extends Component {
 
         const classes = [
             'sk-notification',
-            'sk-notification-error',
-            message && (message.length > 50) && 'long-text'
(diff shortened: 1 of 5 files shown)


chloepouprom:
Nice approach! However, the `sk-footer` has a `flex`: https://github.com/smooch/smooch-js/blob/91f53a14fd8e85c5b9da525bdf3f04a4b063739c/src/stylesheets/footer.less#L12
Because of this, the footer keeps changing height when the error notification appears/disappears.

Not sure if we can get rid of it in the footer. 😐

PR #336, opened 2016-08-02
Prevent intro height calcuation when closing

lemieux:
Long story short :  since the widget is scaling down to disappear, it was triggering the intro height recalculation, which updated the store, which triggered the conversation component to be updated, which caused the small animation glitch. This will now prevent the intro component to recompute its height when the widget is closed or closing (the widget enters the CLOSED state when starting the animation).

I was also tired by using a boolean to represent 3 possible states, so I made it a proper enum like it should have been from the start.

@alavers @dannytranlx @chloepouprom @spasiu @jugarrit

### DIFF
--- a/src/js/components/chat-input.jsx
+++ b/src/js/components/chat-input.jsx
@@ -93,24 +93,22 @@ export class ChatInputComponent extends Component {
             inputContainerClasses.push('no-upload');
         }
 
-        return (
-            <div id='sk-footer'>
-                { imageUploadButton }
-                <form onSubmit={ this.onSendMessage }
-                      action='#'>
-                    <div className={ inputContainerClasses.join(' ') }>
-                        <input ref='input'
-                               placeholder={ ui.text.inputPlaceholder }
-                               className='input message-input'
-                               onChange={ this.onChange }
-                               onFocus={ this.onFocus }
-                               value={ this.state.text }
-                               title={ ui.text.sendButtonText }></input>
-                    </div>
-                </form>
-                { sendButton }
-            </div>
-            );
+        return <div id='sk-footer'>
+                   { imageUploadButton }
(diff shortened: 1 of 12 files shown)


src/js/components/chat-input.jsx
```diff
@@ -93,24 +93,22 @@ export class ChatInputComponent extends Component {
...
-                               onChange={ this.onChange }
-                               onFocus={ this.onFocus }
-                               value={ this.state.text }
-                               title={ ui.text.sendButtonText }></input>
-                    </div>
-                </form>
-                { sendButton }
-            </div>
-            );
+        return <div id='sk-footer'>
```
lemieux: This change is completely unrelated :| ... I was bothered by the `()` around the markup... it's the only change there.

src/stylesheets/mixins.less
```diff
@@ -41,3 +41,8 @@
         background-color: rgba(210, 210, 210, 0.97);
     }
 }
+
+.flex(@value) {
```
lemieux: Hmm, kind of unrelated :(, but I'll use it in another branch

src/js/components/introduction.jsx
```diff
@@ -77,10 +81,12 @@ export class IntroductionComponent extends Component {
     }
 }
 
-export const Introduction = connect(({appState: {introHeight}}) => {
+export const Introduction = connect(({appState: {introHeight, widgetState}}) => {
+    console.log(widgetState, introHeight)
```
chloepouprom: console.log

mspensieri:
👍

chloepouprom:
👍

src/stylesheets/mixins.less
dannytranlx: Jesus 😂

PR #337, opened 2016-08-02
Add new class for appear and close animation

chloepouprom:
This prevents the widget from opening/closing unnecessarily on window resize. I followed @jpjoyal's proposed approach and disabled the animation on window resize.

The animations are now all in the `sk-animation` class. This class gets added only when clicking on the header (and thus toggling the open/close state). When the window resizes, the class gets removed.

I also fixed a few of the media queries. They should now be consistent with theses media queries: https://github.com/smooch/smooch-js/blob/integration/src/stylesheets/main.less#L40

@lemieux @mspensieri @spasiu @dannytranlx

### DIFF
--- a/src/js/components/widget.jsx
+++ b/src/js/components/widget.jsx
@@ -77,7 +77,7 @@ export class WidgetComponent extends Component {
     };
 
     componentWillUnmount = () => {
-        window.addEventListener('resize', this.handleResize);
+        window.removeEventListener('resize', this.handleResize);
     };
 
     render() {
--- a/src/js/components/widget.jsx
+++ b/src/js/components/widget.jsx
@@ -34,6 +34,10 @@ export class WidgetComponent extends Component {
         ui: PropTypes.object.isRequired
     };
 
+    state = {
+        showAnimation: false
+    };
+
     onTouchStart = (e) => {
         resetUnreadCount();
         // the behavior is problematic only on iOS devices
@@ -50,6 +54,11 @@ export class WidgetComponent extends Component {
 
     onClick = () => {
         resetUnreadCount();
+        this.setState({showAnimation: true});
+    };
+
+    handleResize = () => {
+        this.setState({showAnimation: false});
     };
 
     onWheel = debounce(() => {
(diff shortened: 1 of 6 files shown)


src/js/components/widget.jsx
```diff
@@ -33,6 +33,10 @@ export class WidgetComponent extends Component {
         ui: PropTypes.object.isRequired
     };
 
+    state = {
+        showAnimation: false
```
lemieux: this goes in the store ;)

lemieux: mainly because you need to handle the case where someone calls `Smooch.open()` and `Smooch.close()`

src/js/components/widget.jsx
```diff
@@ -49,6 +53,11 @@ export class WidgetComponent extends Component {
 
     onClick = () => {
         resetUnreadCount();
+        this.setState({showAnimation: true});
```
lemieux: I would do this in `openWidget` and `closeWidget` function in the app service instead. This would ensure it's done whenever the action is done.

lemieux: You could even not have an action to do it and just have the reducer turn on the flag when the action is dispatched. You would end up with only a `disableAnimation` action.

test/specs/reducers/app-state-reducer.spec.js
```diff
@@ -0,0 +1,18 @@
+import { AppStateReducer } from '../../../src/js/reducers/app-state-reducer';
+import { TOGGLE_WIDGET, OPEN_WIDGET, CLOSE_WIDGET } from '../../../src/js/actions/app-state-actions';
+
+const INITIAL_STATE = AppStateReducer(undefined, {});
```
lemieux: I'd say force `showAnimation` to be false in the initial state, otherwise we might switch the default to true and you wouldn't know your dispatched action is still switching it on.

lemieux: also, no tests for the disable action?

src/js/components/widget.jsx
```diff
@@ -65,6 +71,14 @@ export class WidgetComponent extends Component {
         };
     }
 
+    componentDidMount = () => {
+        window.addEventListener('resize', this.handleResize);
+    };
+
+    componentWillUnmount = () => {
+        window.addEventListener('resize', this.handleResize);
```
mspensieri: should remove the listener here I think

mspensieri:
👍

lemieux:
:+1:

PR #338, opened 2016-08-02
Add 'Send me a text' button

chloepouprom:
Clicking on the 'Send me a text' button (only visible when on computer) will make a call to `/appUsers/:appUserId/integrations/twilio/ping`

I also added some more error handling.

@jugarrit @dannytranlx @alavers @lemieux

### DIFF
--- a/src/js/components/channels/twilio-channel-content.jsx
+++ b/src/js/components/channels/twilio-channel-content.jsx
@@ -25,15 +25,8 @@ export class TwilioChannelContentComponent extends Component {
 
     handleInputChange = (telNumber) => {
         updateTwilioAttributes({
-            appUserNumber: telNumber
-        });
-    }
-
-    onChangeNumber = () => {
-        updateTwilioAttributes({
-            linkState: 'unlinked',
-            appUserNumber: '',
-            appUserNumberValid: false
+            appUserNumber: telNumber,
+            hasError: false
         });
     }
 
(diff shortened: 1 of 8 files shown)


src/js/services/integrations-service.js
```diff
@@ -63,10 +63,17 @@ export function linkTwilioChannel(userId, data) {
...
             });
         })
-        .catch(() => {
+        .catch((e) => {
+            const {response: {status}} = e;
+            let errorMessage;
+            if (status === 429) {
+                errorMessage = 'A link was attempted recently for this phone number. Please wait before trying again.';
+            } else {
+                errorMessage = 'We were unable to communicate with this number. Please enter a different one.';
```
jugarrit: If a `400` is returned this is the appropriate error, but if it's `500` maybe a try again might be good

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -97,10 +93,10 @@ export class TwilioChannelContentComponent extends Component {
                                  </div>;
 
         const sendTextUrl = `sms://${phoneNumber}`;
-        const linkedComponentButton = isMobile.phone ? <button className='btn btn-sk-primary twilio-linking'
-                                                               onClick={ this.onStartTexting }>
-                                                           Start Texting
-                                                       </button> :
+        const linkedComponentButton = isMobile.phone ? <a href={ sendTextUrl }>
```
lemieux: why is there a button inside of a `a`?

lemieux:
other than the button thing :+1:

mspensieri:
All these new strings should live in the string constants dictionary, and should be customizable using the `customText` init property

mspensieri:
👍

jugarrit:
:+1:

PR #342, opened 2016-08-04
Preload country list to speed up load times

jugarrit:
@chloepouprom @lemieux @mspensieri @dannytranlx 

The country dropdown would take too long to open when it was clicked. This will speed it up.

### DIFF
--- a/src/js/lib/react-telephone-input.jsx
+++ b/src/js/lib/react-telephone-input.jsx
@@ -83,6 +83,7 @@ export class ReactTelephoneInput extends Component {
             selectedCountry: selectedCountryGuess,
             highlightCountryIndex: selectedCountryGuessIndex,
             formattedNumber: formattedNumber,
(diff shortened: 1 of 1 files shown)


lemieux:
Does it speed it up?

jugarrit:
AHahah yes it seems so.
I'll try to make a gif comparison

jugarrit:
This probably does nothing.

src/js/lib/react-telephone-input.jsx
```diff
@@ -454,7 +457,7 @@ export class ReactTelephoneInput extends Component {
             }, function() {
                 self.scrollTo(self.getElement(self.state.highlightCountryIndex), true);
             });
-        }
```
dannytranlx: This was needed.

lemieux:
Yeah it probably does nothing

PR #344, opened 2016-08-04
Fix background scroll in iOS

chloepouprom:
When there are no messages, background should not scroll.

@mspensieri @lemieux @dannytranlx

### DIFF
--- a/src/js/components/widget.jsx
+++ b/src/js/components/widget.jsx
@@ -135,7 +135,7 @@ export class WidgetComponent extends Component {
         return <div>
                    <div id='sk-container'
                         className={ classNames.join(' ') }
-                        onTouchStart={ this.onTouchStart }
+                        onTouchStart={ this.onTouchStart.bind(this) }
                         onClick={ this.onClick }
                         onWheel={ this.onWheel }>
                        <MessageIndicator />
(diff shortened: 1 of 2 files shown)


src/js/components/widget.jsx
```diff
@@ -118,7 +118,7 @@ export class WidgetComponent extends Component {
         return <div>
                    <div id='sk-container'
                         className={ classNames.join(' ') }
-                        onTouchStart={ this.onTouchStart }
+                        onTouchStart={ this.onTouchStart.bind(this) }
```
lemieux: you shouldn't need to do that since it was defined with `=>`

lemieux:
Still works when there are messages?

src/js/components/widget.jsx
chloepouprom: While debugging on console, I noticed that `this` was undefined :|

lemieux: was it throwing errors? because `this` in the console is tricky because of transpilation... sometimes in the console it's `_this`

chloepouprom: Aaaaah okay. There were no errors.

chloepouprom:
Yep, still works with messages!

jugarrit:
👍

lemieux:
:+1:

PR #345, opened 2016-08-04
Set error for SMS linking from config, use retry-after

jugarrit:
@chloepouprom @lemieux @dannytranlx @mspensieri 

Error was hardcoded previously, now using config values.
Also making use of retry-after returned from server.


chloepouprom:
Oops, I think I missed those during rebase. 👍

mspensieri:
👍

lemieux:
:+1:

dannytranlx:
👍

PR #347, opened 2016-08-04
Fix embedded mode

chloepouprom:
With `position: absolute`, the element is positioned relative to its first positioned (not static) ancestor element. 🎉 

@lemieux @dannytranlx @mspensieri @Go-Fred

### DIFF
--- a/src/stylesheets/main.less
+++ b/src/stylesheets/main.less
@@ -118,6 +118,7 @@
     @import "reset.less";
     .container();
     height: 100%;
+    position: absolute;
     #sk-wrapper {
         background: @background-color;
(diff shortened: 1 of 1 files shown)


lemieux:
:+1:

chloepouprom:
`sk-container` has `height: 100%`. By setting `position: absolute`, we ensure that the height is with respect to the parent container (instead of the whole page).

lemieux:
image

dannytranlx:
👍

mspensieri:
👍

sotojuan:
When can this feature be put in a release? We really need it!

PR #352, opened 2016-08-05
Display messenger button on iOS7

chloepouprom:
The messenger button wasn't showing up on iOS7. Adding webkit to animations and transforms fixes this problem.

@lemieux @mspensieri @dannytranlx

### DIFF
--- a/src/stylesheets/animations.less
+++ b/src/stylesheets/animations.less
@@ -119,21 +119,27 @@
         }
 
         &.sk-animation {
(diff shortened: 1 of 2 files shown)


lemieux:
I'm surprised it works without having to set webkit keyframes too.

:+1:

chloepouprom:
I'll add the webkit keyframes too just to be safe

dannytranlx:
👍

PR #353, opened 2016-08-05
Fix IE11 button animation

lemieux:
The problem was that after finishing the animation, the button would go back to the 0% keyframe (which hides it) in IE11. Setting animation-fill-mode to `both` fixed that. I tested it IE10, IE11, Chrome, Android and iOS.

@alavers  @chloepouprom @dannytranlx @mspensieri


chloepouprom:
👍

dannytranlx:
:

PR #356, opened 2016-08-09
Update UI after link:cancelled event

chloepouprom:
@jugarrit @lemieux @mspensieri @dannytranlx

### DIFF
--- a/src/js/components/channels/twilio-channel-content.jsx
+++ b/src/js/components/channels/twilio-channel-content.jsx
@@ -26,8 +26,7 @@ export class TwilioChannelContentComponent extends Component {
 
     handleInputChange = (telNumber) => {
         updateTwilioAttributes({
-            appUserNumber: telNumber,
-            hasError: false
+            appUserNumber: telNumber
         });
     }
 
@@ -43,7 +42,8 @@ export class TwilioChannelContentComponent extends Component {
 
     onNumberValid = () => {
         updateTwilioAttributes({
-            appUserNumberValid: true
+            appUserNumberValid: true,
+            hasError: false
         });
     }
 
--- a/src/js/services/integrations-service.js
+++ b/src/js/services/integrations-service.js
@@ -133,3 +133,16 @@ export function pingTwilioChannel(userId) {
             });
         });
 }
+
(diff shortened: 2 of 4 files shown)


src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -26,8 +26,7 @@ export class TwilioChannelContentComponent extends Component {
 
     handleInputChange = (telNumber) => {
         updateTwilioAttributes({
-            appUserNumber: telNumber,
-            hasError: false
```
chloepouprom: I reset the `hasError` flag in `onNumberValid` instead. 

When updating the UI following the `link:cancelled` event, `handleInputChange` would get triggered and would reset the `hasError` flag, thus never rendering the message. 😐

src/js/services/integrations-service.js
```diff
@@ -133,3 +133,16 @@ export function pingTwilioChannel(userId) {
...
         });
 }
+
+export function cancelTwilioLink() {
+    const {user: {pendingClients}, integrations: {twilio: {appUserNumber}}, ui: {text: {smsLinkCancelled}}} = store.getState();
+
+    store.dispatch(updateUser({
+        pendingClients: pendingClients.filter((pendingClient) => pendingClient.platform !== 'twilio')
+    }));
+    updateTwilioAttributes({
```
lemieux: I'd put a space between those

src/js/utils/faye.js
```diff
@@ -124,6 +125,8 @@ export function handleUserSubscription({appUser, event}) {
                 }, ANIMATION_TIMINGS.PAGE_TRANSITION);
             }, ANIMATION_TIMINGS.PAGE_TRANSITION);
         }
+    } else if (event.type === 'link:cancelled') {
+        return cancelTwilioLink();
```
lemieux: I would check the platform before saying it's twilio being cancelled. Sure we only support that for now, but we don't know what the future holds for us and on which version of the widget people will be stuck when we'll roll out new integrations.

lemieux:
:+1:

jugarrit:
👍

PR #360, opened 2016-08-09
Fix for wrong isOpened behavior

lemieux:
Small fix for the isOpened function. Forgot to remove the `!!` when replacing the boolean with proper states.

I'll merge right away and deploy since this is a pretty small targeted fix.

@chloepouprom @dannytranlx @jugarrit


lemieux:
Will fix #360

PR #366, opened 2016-08-12
Hide channel icon and name to avoid setting header disappearing

jugarrit:
@lemieux @alavers @mspensieri @chloepouprom 

When the phone number input is selected, the window shrinks, causing Chrome for Android to push the header up above the URL bar. Hiding the channel icon and header for extremely small heights fixes this issue.

image

### DIFF
--- a/src/stylesheets/channels.less
+++ b/src/stylesheets/channels.less
@@ -30,7 +30,7 @@
 
     .channel-header {
         .channel-icon, .channel-name {
-            @media (max-height: 400px) {
+            @media (max-height: 400px) and (max-width: @screen-sm-min) {
                 display: none;
             }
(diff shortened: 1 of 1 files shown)


src/stylesheets/channels.less
```diff
@@ -29,6 +29,12 @@
     }
 
     .channel-header {
+        .channel-icon, .channel-name {
+            @media (max-height: 400px) {
```
chloepouprom: Does the setting header only disappear on mobile screens? If so I would also add a `(max-width: @screen-sm-min)` to the media query.

mspensieri:
👍

chloepouprom:
👍

PR #374, opened 2016-08-17
Add pagination in web widget

chloepouprom:
Using the new API endpoint, only 100 messages are fetched at time. When scrolling up, older messages will be fetched.

pagination_1

pagination_3 pagination_2

In order to get the render to work well, I had to make a few changes to the `Introduction` component. Instead of calculating its height on every `componentWillUpdate`, I only calculate it on `componentDidMount` and on a resize event. This fixes some of the weird animation glitches we had observed before.

A few things to note on the scroll behavior:
- An appUser sending a message will cause the conversation to scroll the bottom.
- If the appUser has scrolled the conversation up (towards older messages), and an appMaker sends a message, then the conversation will not scroll to the bottom. 
- If the appUser is at the bottom of the conversation (newer messages), and an appMaker sends a message, then the conversation will scroll to the bottom.
- Scrolling to the 'top' of the conversation will fetch older messages. 
- Older messages can also be fetched by clicking on the `Load more` link. This link shows up when you've scrolled up (but not all the way to the top!)
- When older messages get fetched, the messages stay in the same position they originally were (the first screenshot shows this well 😁 )

@mspensieri @dannytranlx @alavers @lemieux

### DIFF
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -70,9 +70,10 @@ export class ConversationComponent extends Component {
 
         // If top of Conversation component is reached, we need to fetch older messages
         const node = findDOMNode(this);
-        if(node.scrollTop === 0) {
+        if (node.scrollTop === 0) {
             this.fetchHistory();
-        } else if(shouldScrollToBottom) {
+        } else if (shouldScrollToBottom) {
+            // Once we've started scrolling, we don't want the default behavior to force the scroll to the bottom afterwards
             dispatch(setShouldScrollToBottom(false));
         }
     };
@@ -81,7 +82,7 @@ export class ConversationComponent extends Component {
         const {dispatch, hasMoreMessages, isFetchingMoreMessages, messages} = this.props;
 
         const node = findDOMNode(this);
-        if(hasMoreMessages && !isFetchingMoreMessages) {
+        if (hasMoreMessages && !isFetchingMoreMessages) {
             // make sure the last message is one from the server, otherwise it doesn't need to scroll to previous first message
             if (messages.length > 0 && messages[messages.length - 1]._id) {
                 this._lastTopMessageId = messages[0]._id;
@@ -113,7 +114,7 @@ export class ConversationComponent extends Component {
     };
 
     scrollToPreviousFirstMessage = () => {
-        if(this._lastTopMessageNodePosition && !this._isScrolling) {
+        if (this._lastTopMessageNodePosition && !this._isScrolling) {
             const container = findDOMNode(this);
             const node = this._lastTopMessageNode;
             this._isScrolling = true;
@@ -139,8 +140,8 @@ export class ConversationComponent extends Component {
             this._forceScrollToBottom = true;
         }
 
-        // Check for new appMaker messages
-        const isAppMakerMessage = newMessages.length - currentMessages.length === 1 ? newMessages.slice(-1)[0].role === 'appMaker' : false;
+        // Check for new appMaker (and whisper) messages
+        const isAppMakerMessage = newMessages.length - currentMessages.length === 1 ? newMessages.slice(-1)[0].role !== 'appUser' : false;
         if (isAppMakerMessage && !isFetchingMoreMessages) {
             const container = findDOMNode(this);
             const appMakerMessageBottom = this._lastMessageNode.getBoundingClientRect().bottom;
@@ -172,14 +173,6 @@ export class ConversationComponent extends Component {
 
     componentWillUnmount() {
         this.scrollTimeouts.forEach(clearTimeout);
-        delete this._lastTopMessageNode;
-        delete this._lastTopMessageId;
-        delete this._lastTopMessageNodePosition;
-        delete this._topMessageNode;
-        delete this._forceScrollToBottom;
-        delete this._isScrolling;
-        delete this._lastMessageNode;
-        delete this._lastMessageId;
     }
 
     render() {
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -75,10 +75,15 @@ export class ConversationComponent extends Component {
     };
 
     fetchHistory = () => {
-        const {dispatch, hasMoreMessages, isFetchingMoreMessages} = this.props;
+        const {dispatch, hasMoreMessages, isFetchingMoreMessages, messages} = this.props;
 
         const node = findDOMNode(this);
         if(hasMoreMessages && !isFetchingMoreMessages) {
+            // make sure the last message is one from the server, otherwise it doesn't need to scroll to previous first message
+            if (messages.length > 0 && messages[messages.length - 1]._id) {
+                this._lastTopMessageId = messages[0]._id;
+            }
+            
             const top = getTop(this._topMessageNode, node);
             this._lastTopMessageNodePosition = top - node.scrollTop;
             dispatch(setFetchingMoreMessages(true));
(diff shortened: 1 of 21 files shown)


src/js/components/conversation.jsx
```diff
@@ -54,34 +65,145 @@ export class ConversationComponent extends Component {
...
 
+    onScroll = () => {
+        const {dispatch, shouldScrollToBottom} = this.props;
+
+        // If top of Conversation component is reached, we need to fetch older messages
+        const node = findDOMNode(this);
+        if(node.scrollTop === 0) {
+            this.fetchHistory();
+        } else if(shouldScrollToBottom) {
+            dispatch(setShouldScrollToBottom(false));
```
lemieux: maybe a small comment about why we set it to false

src/js/services/conversation-service.js
```diff
@@ -235,3 +246,28 @@ export function postPostback(actionId) {
         store.dispatch(showErrorNotification(store.getState().ui.text.actionPostbackError));
     });
 }
+
+
+export function fetchMoreMessages() {
+    const {conversation: {hasMoreMessages, messages}} = store.getState();
+
+    // TODO : add independent guard against parallel fetching
```
lemieux: still todo ;)

lemieux:
Aside the TODO part I'll :+1: the part I didn't write

mspensieri:
I don't agree with this bullet point

> Scrolling to the 'top' of the conversation will fetch older messages.

IMO we should only load old messages when the user taps the "load more" link

src/js/components/conversation.jsx
```diff
@@ -54,34 +65,145 @@ export class ConversationComponent extends Component {
...
+        if(hasMoreMessages && !isFetchingMoreMessages) {
+            // make sure the last message is one from the server, otherwise it doesn't need to scroll to previous first message
+            if (messages.length > 0 && messages[messages.length - 1]._id) {
+                this._lastTopMessageId = messages[0]._id;
+            }
+            
+            const top = getTop(this._topMessageNode, node);
+            this._lastTopMessageNodePosition = top - node.scrollTop;
+            dispatch(setFetchingMoreMessages(true));
+            setTimeout(() => {
```
mspensieri: why set timeout?

src/js/components/conversation.jsx
```diff
@@ -8,11 +8,17 @@ import { ConnectNotification } from './connect-notification';
 import { logo, logo2x } from '../constants/assets';
```
mspensieri: needs some formatting in this file

src/js/components/conversation.jsx
```diff
@@ -54,34 +65,145 @@ export class ConversationComponent extends Component {
...
+    componentWillUpdate(nextProps) {
+        const {introHeight: currentIntroHeight, messages: currentMessages, isFetchingMoreMessages} = this.props;
+        const {introHeight: newIntroHeight, messages: newMessages} = nextProps;
+
+        if (currentIntroHeight !== newIntroHeight) {
+            this._forceScrollToBottom = true;
+        }
+
+        // Check for new appMaker messages
+        const isAppMakerMessage = newMessages.length - currentMessages.length === 1 ? newMessages.slice(-1)[0].role === 'appMaker' : false;
```
mspensieri: `.role !== 'appUser'` is a better check. A message with role `whisper` is still technically an app maker message

src/js/components/conversation.jsx
```diff
@@ -54,34 +65,145 @@ export class ConversationComponent extends Component {
...
+        if (this.props.isFetchingMoreMessages) {
+            this.scrollToPreviousFirstMessage();
+        } else {
+            this.scrollToBottom();
+        }
     }
 
     componentWillUnmount() {
         this.scrollTimeouts.forEach(clearTimeout);
+        delete this._lastTopMessageNode;
```
mspensieri: Why is this necessary? Doesn't the component get recycled after it unmounts?

src/js/services/conversation-service.js
chloepouprom: Hmmm I guess I can just check the `appState` to see if `isFetchingMoreMessages` is true. Since we set it right before calling it.

chloepouprom:
> IMO we should only load old messages when the user taps the "load more" link

We do this so that we can have infinite scroll

src/js/components/conversation.jsx
chloepouprom: True. I'll remove

src/js/components/conversation.jsx
chloepouprom: I think this was originally so that the transition from the `Load more` to `Retrieving history` text appears less choppy. But now, looks like it doesn't actually make a difference... I'll remove this

src/js/services/conversation-service.js
lemieux: Can't do that since you would effectively  be able to call it multiple time
in a row and it would be true... Just need a flag independent of the UI one

src/js/components/conversation.jsx
lemieux: You usually use setTimeout when you need to compute sizes because the
function in setTimeout will be executed when the html is effectively
flushed to the dom so it would reflect the real sizes with css and
everything

lemieux:
Yes that was a requirement of the feature... The load more button is
actually a workaround to prevent people from guessing if there are more
messages or not if they didn't go all the way to the top.

mspensieri:
infinite scroll is kind of annoying when you have a banner at the top that you might be trying to interact with... "I would like to link messenger. Oops, scrolled 1 pixel too far, it loaded more messages. Guess I'll try again. Oops, scrolled 1 pixel too far"

src/js/components/conversation.jsx
```diff
@@ -54,34 +65,145 @@ export class ConversationComponent extends Component {
...
+            }, 400);
+        }
+    };
+
     scrollToBottom = () => {
-        const timeout = setTimeout(() => {
+        const {shouldScrollToBottom} = this.props;
+        if(!this._isScrolling && (shouldScrollToBottom || this._forceScrollToBottom)) {
+            this._isScrolling = true;
+            const timeout = setTimeout(() => {
```
alavers: There doesn't seem any need for this function to be a closure like this. You could have what `setTimeout` is scheduling instead be a class method called `scrollToBottom`, then the outer method would be `scrollToBottomIfNeeded`

lemieux:
Yeah we thought about that and since we have the settings page too we felt
it was fine.

However, now that I think more about it, I don't like the infinite scroll
because we have stuff at the top of the conversation unlike slack.

It think we should make a case to Product about it and remove the infinite
scrolling.

On Aug 18, 2016 16:23, "Mike Spensieri" <notifications@github.com> wrote:

infinite scroll is kind of annoying when you have a banner at the top that
you might be trying to interact with... "I would like to link messenger.
Oops, scrolled 1 pixel too far, it loaded more messages. Guess I'll try
again. Oops, scrolled 1 pixel too far"

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/smooch/smooch-js/pull/374#issuecomment-240844942>, or mute
the thread

src/js/components/conversation.jsx
lemieux: The reason why it's in a closure within setTimeout is because you want to have the `clientHeight` and `scrollHeight` when the content is flushed to the DOM. This was a valid thing to do back when we used Marionette and I assumed it still was necessary with React. Maybe it's not... we should probably test that out. 

But you are right that this could be made a class method if `setTimeout` is still a thing to do.

PR #375, opened 2016-08-19
Resolve warning for 'flood-color' should be actually 'floodColor'

brene:
Currently, if you use this component, it works but the console keeps throwing warnings. Let's keep a 0 warning policy 👍


lemieux:
Hey there, thanks for the PR :) 

I'll merge this and since it's not blocking anything, it will just go out with the next release.

Thanks again for the contribution :) I'll also close your related issue.

PR #378, opened 2016-08-22
Add pagination in web widget

chloepouprom:
I closed the previous pull request https://github.com/smooch/smooch-js/pull/374 since this would have otherwise gotten confusing with all the comments.

Anyway, this pretty much does the same as the previous PR except that now the Introduction component is only present at the very beginning of the conversation.

pagination_intro

@lemieux @alavers @mspensieri

### DIFF
--- a/src/js/actions/conversation-actions.js
+++ b/src/js/actions/conversation-actions.js
@@ -7,6 +7,7 @@ export const SET_CONVERSATION = 'SET_CONVERSATION';
 export const SET_MESSAGES = 'SET_MESSAGES';
 export const RESET_UNREAD_COUNT = 'RESET_UNREAD_COUNT';
 export const INCREMENT_UNREAD_COUNT = 'INCREMENT_UNREAD_COUNT';
+export const SET_FETCHING_MORE_MESSAGES_FROM_SERVER = 'SET_FETCHING_MORE_MESSAGES_FROM_SERVER';
 
 export function resetConversation() {
     return {
@@ -69,5 +70,12 @@ export function incrementUnreadCount() {
 export function resetUnreadCount() {
     return {
         type: RESET_UNREAD_COUNT
+    };
+}
+
+export function setFetchingMoreMessagesFromServer(value) {
+    return {
+        type: SET_FETCHING_MORE_MESSAGES_FROM_SERVER,
+        value
     };
 }
--- a/src/js/actions/conversation-actions.js
+++ b/src/js/actions/conversation-actions.js
@@ -1,8 +1,10 @@
 export const ADD_MESSAGE = 'ADD_MESSAGE';
+export const ADD_MESSAGES = 'ADD_MESSAGES';
 export const REPLACE_MESSAGE = 'REPLACE_MESSAGE';
 export const REMOVE_MESSAGE = 'REMOVE_MESSAGE';
 export const RESET_CONVERSATION = 'RESET_CONVERSATION';
 export const SET_CONVERSATION = 'SET_CONVERSATION';
+export const SET_MESSAGES = 'SET_MESSAGES';
 export const RESET_UNREAD_COUNT = 'RESET_UNREAD_COUNT';
 export const INCREMENT_UNREAD_COUNT = 'INCREMENT_UNREAD_COUNT';
 
@@ -19,12 +21,27 @@ export function setConversation(props) {
     };
 }
 
+export function setMessages(messages) {
+    return {
+        type: SET_MESSAGES,
+        messages
+    };
+}
+
 export function addMessage(props) {
     return {
         type: ADD_MESSAGE,
         message: Object.assign({
             actions: []
         }, props)
+    };
+}
+
+export function addMessages(messages, append = true) {
+    return {
+        type: ADD_MESSAGES,
+        messages,
+        append
     };
 }
 
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -70,9 +70,10 @@ export class ConversationComponent extends Component {
 
         // If top of Conversation component is reached, we need to fetch older messages
         const node = findDOMNode(this);
-        if(node.scrollTop === 0) {
+        if (node.scrollTop === 0) {
             this.fetchHistory();
-        } else if(shouldScrollToBottom) {
+        } else if (shouldScrollToBottom) {
+            // Once we've started scrolling, we don't want the default behavior to force the scroll to the bottom afterwards
             dispatch(setShouldScrollToBottom(false));
         }
     };
@@ -81,7 +82,7 @@ export class ConversationComponent extends Component {
         const {dispatch, hasMoreMessages, isFetchingMoreMessages, messages} = this.props;
 
         const node = findDOMNode(this);
-        if(hasMoreMessages && !isFetchingMoreMessages) {
+        if (hasMoreMessages && !isFetchingMoreMessages) {
             // make sure the last message is one from the server, otherwise it doesn't need to scroll to previous first message
             if (messages.length > 0 && messages[messages.length - 1]._id) {
                 this._lastTopMessageId = messages[0]._id;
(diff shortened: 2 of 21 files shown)


kvgauthier:
@chloepouprom Can you add a couple of pixels of margin above the loading text. Around 24 pixels.

src/js/components/introduction.jsx
```diff
@@ -26,35 +24,31 @@ export class IntroductionComponent extends Component {
...
-
-    componentWillUpdate() {
-        this._debounceHeightCalculation();
+        // Height of Introduction component will be computed on render and on resize only
+        // Make sure Introduction component has fully rendered before computing height
+        setTimeout(() => {
+            this.calculateIntroHeight();
+        }, 200);
+
+        window.addEventListener('resize', () => {
```
lemieux: this should be debounced on resize otherwise it will be triggered and executed a lot of times during the resize.

src/js/actions/conversation-actions.js
```diff
@@ -54,3 +72,10 @@ export function resetUnreadCount() {
         type: RESET_UNREAD_COUNT
     };
 }
+
+export function setFetchingMoreMessagesFromServer(value) {
```
alavers: Do we need this distinct event? Could we reuse the `setFetchingMoreMessages ` app state action?

chloepouprom: This is to make sure there aren't multiple fetch requests happening at the same time
https://github.com/smooch/smooch-js/pull/374#discussion-diff-75224489

src/js/components/conversation.jsx
```diff
@@ -54,34 +66,146 @@ export class ConversationComponent extends Component {
...
+            // Once we've started scrolling, we don't want the default behavior to force the scroll to the bottom afterwards
+            dispatch(setShouldScrollToBottom(false));
+        }
+    };
+
+    fetchHistory = () => {
+        const {dispatch, hasMoreMessages, isFetchingMoreMessages, messages} = this.props;
+
+        const node = findDOMNode(this);
+        if (hasMoreMessages && !isFetchingMoreMessages) {
```
alavers: I think instead of making this if statement check inside fetchHistory, it should be `onScroll` who decides whether or not it should call `fetchHistory`.

There are already such checks deciding if fetchHistory gets called in the `render()` method below.

src/js/components/conversation.jsx
```diff
@@ -54,34 +66,146 @@ export class ConversationComponent extends Component {
...
+        // On component render, force scroll to bottom, or else conversation will
+        // find itself at a random spot
         this.scrollToBottom();
     }
 
     componentDidUpdate() {
-        this.scrollToBottom();
+        if (this.props.isFetchingMoreMessages) {
+            if (this.props.hasMoreMessages) {
+                this.scrollToPreviousFirstMessage();
```
alavers: To make this logc clearer `scrollToPreviousFirstMessage` ought to be checking `this.props.hasMoreMessages` itself

src/js/components/conversation.jsx
```diff
@@ -54,34 +66,146 @@ export class ConversationComponent extends Component {
...
+    onScroll = () => {
+        const {dispatch, shouldScrollToBottom} = this.props;
+
+        // If top of Conversation component is reached, we need to fetch older messages
+        const node = findDOMNode(this);
+        if (node.scrollTop === 0) {
+            this.fetchHistory();
+        } else if (shouldScrollToBottom) {
+            // Once we've started scrolling, we don't want the default behavior to force the scroll to the bottom afterwards
+            dispatch(setShouldScrollToBottom(false));
```
alavers: How rapidly does `onScroll` get fired? Should we be throttling this event dispatch?

src/js/components/conversation.jsx
chloepouprom: Hmm.. `onScroll` gets fired often so I thought it would be better to keep this logic in `fetchHistory`

src/js/components/conversation.jsx
chloepouprom: `onScroll` get debounced and fires every 200ms on non-mobile devices

I had to do it this way because the mobile scroll was behaving really weirdly.

src/js/components/conversation.jsx
```diff
@@ -54,34 +66,146 @@ export class ConversationComponent extends Component {
...
+
+                this.scrollTimeouts.push(timeout);
+            }
+        }
+        this._lastTopMessageNode = undefined;
+        
     };
 
+    componentWillUpdate(nextProps) {
+        const {messages: currentMessages, isFetchingMoreMessages} = this.props;
```
alavers: Is there a reason why we do this in `componentWillUpdate`? It feels to me like inner workings of `scrollToBottom` are leaking out into `componentWillUpdate`, and we could perhaps instead have a `is there an unseen appmaker message` method that `scrollToBottom` calls opportunistically, instead of having this run for every `componentWillUpdate`.

src/js/components/introduction.jsx
```diff
@@ -26,35 +25,30 @@ export class IntroductionComponent extends Component {
...
+        this._debounceClientHeightCalculation = debounce(this.calculateIntroHeight.bind(this), 150);
     }
 
     componentDidMount() {
-        setTimeout(() => this.calculateIntroHeight());
-        window.addEventListener('resize', this._debounceHeightCalculation);
-    }
+        // Height of Introduction component will be computed on render and on resize only
+        // Make sure Introduction component has fully rendered before computing height
+        setTimeout(() => {
```
alavers: is there any way to synchronize with intro rendering completion that doesn't rely on a setTimeout like this? Could we synchronize on `componentDidUpdate` for example?

src/js/components/conversation.jsx
```diff
@@ -54,34 +66,146 @@ export class ConversationComponent extends Component {
...
+        if (hasMoreMessages && !isFetchingMoreMessages) {
+            // make sure the last message is one from the server, otherwise it doesn't need to scroll to previous first message
+            if (messages.length > 0 && messages[messages.length - 1]._id) {
+                this._lastTopMessageId = messages[0]._id;
+            }
+            
+            const top = getTop(this._topMessageNode, node);
+            this._lastTopMessageNodePosition = top - node.scrollTop;
+            dispatch(setFetchingMoreMessages(true));
+            setTimeout(() => {
```
alavers: Could also add a comment to explain why this setTimeout is necessary

alavers:
Rest looks good 👍

src/js/components/conversation.jsx
alavers: gotcha 👍

src/js/components/conversation.jsx
alavers: Oh, yeah for sure, I'm only suggesting that the `if (hasMoreMessages && !isFetchingMoreMessages) {` statement so that it wraps the call `onScroll` makes to fetchHistory instead of doing it inside fetchHistory itself.

Also, the `const node = findDOMNode(this);` call you make isn't necessary if the if statement is false :)

src/js/components/conversation.jsx
chloepouprom: I put this in `componentWillUpdate` because I wanted to check for a newly received message. `componentWillUpdate` gives me access to the new props and the old props. I compare the length of the messages in the props to see if a new message was sent.

alavers:
👍

src/js/components/conversation.jsx
```diff
@@ -22,6 +29,11 @@ export class ConversationComponent extends Component {
         errorNotificationMessage: PropTypes.string
     };
 
+    constructor(...args) {
```
lemieux: don't need the constructor to do this, you can use a class property... just do  `debounceOnScroll = debounce(this.onScroll.bind(this), 200);`.... like `scrollTimeouts` just below

lemieux:
:+1:

PR #379, opened 2016-08-23
Prevent chat input from inheriting CSS from page

chloepouprom:
In some cases, the chat input would take on the CSS of the page it was on, causing the style to clash with the current Smooch style. To prevent this from happening, force the Smooch chat input to inherit from its parent element.

If you want to customize the chat input, you can edit the CSS: `#sk-holder #sk-container .input`

@jugarrit @dannytranlx @mspensieri

### DIFF
--- a/src/stylesheets/main.less
+++ b/src/stylesheets/main.less
@@ -103,6 +103,7 @@
             background-color: white;
             border: 1px solid white;
             padding: 0px;
+            font: inherit;
         }
 
         .content-wrapper {


mspensieri:
👍

dannytranlx:
Just confirming, that would cover `font-style` and `font-weight` also?

👍

chloepouprom:
@dannytranlx Yes http://www.w3schools.com/cssref/pr_font_font.asp

PR #380, opened 2016-08-23
Fix error notification padding

chloepouprom:
I removed the changing top padding on the Introduction component and I removed the height on the Notification element (it's defined in the animation). 

This works for both one-line and two-line messages.

error-notification-short-message
error-notif-long-message

@dannytranlx @mspensieri @spasiu @lemieux

### DIFF
--- a/src/stylesheets/notification.less
+++ b/src/stylesheets/notification.less
@@ -6,13 +6,7 @@
 
     .sk-notification {
         overflow: hidden;
-
-        height: @notification-height;
         width: 100%;
-
-        &.long-text {
-            height: @long-notification-height;
-        }
 
         border-top: 1px solid rgba(0, 0, 0, .1);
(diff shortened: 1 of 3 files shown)


src/stylesheets/notification.less
```diff
@@ -6,14 +6,8 @@
 
     .sk-notification {
         overflow: hidden;
-
-        height: @notification-height;
```
mspensieri: Are `@notification-height` and `@long-notification-height` still used somewhere? Can we delete those vars?

chloepouprom: `@notification-height` is still used, but `@long-notification-height` isn't. I'll remove it.

mspensieri:
👍

dannytranlx:
👍

PR #381, opened 2016-08-25
Use new API endpoints

chloepouprom:
Use new endpoints for sending message and uploading image https://github.com/smooch/smooch-core-js/pull/45

@mspensieri @lemieux @dannytranlx @jugarrit

### DIFF
--- a/src/js/services/conversation-service.js
+++ b/src/js/services/conversation-service.js
@@ -89,7 +89,7 @@ export function sendMessage(text) {
 
         const {user} = store.getState();
 
-        return core().conversations.sendMessage(getUserId(), message).then((response) => {
+        return core().appUsers.sendMessage(getUserId(), message).then((response) => {
             if (!user.conversationStarted) {
                 // use setConversation to set the conversation id in the store
                 store.dispatch(setConversation(response.conversation));
@@ -137,7 +137,7 @@ export function uploadImage(file) {
             const {user} = store.getState();
             const blob = getBlobFromDataUrl(dataUrl);
 
-            return core().conversations.uploadImage(getUserId(), blob, {
+            return core().appUsers.uploadImage(getUserId(), blob, {
                 role: 'appUser',
                 deviceId: getDeviceId()
             }).then((response) => {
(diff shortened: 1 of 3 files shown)


lemieux:
:+1:

jugarrit:
👍

src/js/services/conversation-service.js
```diff
@@ -89,7 +89,7 @@ export function sendMessage(text) {
 
         const {user} = store.getState();
 
-        return core().conversations.sendMessage(getUserId(), message).then((response) => {
+        return core().appUsers.sendMessage(getUserId(), message).then((response) => {
             if (!user.conversationStarted) {
```
mspensieri: Is the else block of this still relevant for an `else`? Seems like it should execute every time now

src/js/services/conversation-service.js
```diff
@@ -137,7 +137,7 @@ export function uploadImage(file) {
             const {user} = store.getState();
             const blob = getBlobFromDataUrl(dataUrl);
 
-            return core().conversations.uploadImage(getUserId(), blob, {
+            return core().appUsers.uploadImage(getUserId(), blob, {
                 role: 'appUser',
                 deviceId: getDeviceId()
             }).then((response) => {
```
mspensieri: Same question, line 151

src/js/services/conversation-service.js
chloepouprom: There's a different logic for a new conversation and an already started conversation.

mspensieri: But that logic is bogus now isn't it? When you create a conversation, it doesn't include the messages in the response anymore, so we need to update the local message to match the one on the server (which is what I think this `else` is doing)

chloepouprom: This is done through the `SET_MESSAGES` action. This gets dispatched when connecting to Faye.

chloepouprom: Hmm... Actually, I think `SET_MESSAGES` might be firing because of the `handleConnectNotification` function

mspensieri: But that relies on a separate call to the `GET /messages` endpoint, right? So the store is in an inconsistent state until that time - the message is still considered not-uploaded until that call completes and the list is refreshed

mspensieri:
👍

PR #382, opened 2016-08-25
Check for timestamp instead of message id

chloepouprom:
The messages endpoint checks the timestamp instead of id.

@dannytranlx @alavers @mspensieri


mspensieri:
👍

lemieux:
:+1:

Issue #357, opened 2016-08-09
Doesn't work with Webpack 2


schickling:
More and more people are using Webpack 2 and most starter kit are based on it nowadays. Unfortunately your package doesn't work with it.

Could you please either make it compatible or provide a prebuilt file?

lemieux:
You can try using the prebuilt package we have on the CDN at https://cdn.smooch.io/smooch.min.js

schickling:
@lemieux how would I use it via npm?

lemieux:
I guess for now you can't, but you could vendor in that file and alias it with webpack. I'll add the prebuilt package npm suggestion to our backlog.

In the meantime, what is not working with Typescript 2?

schickling:
Oh apologies. I actually meant Webpack 2. (Seems like I cannot control my excitement about Typescript 2 😄

lemieux:
So, same question : what's not working with webpack 2? :)

schickling:
![screen shot 2016-08-09 at 23 36 41](https://cloud.githubusercontent.com/assets/1567498/17534606/3716c7ec-5e8a-11e6-8a15-cdec4ba0ea86.png)

```
 [1751] ./~/websocket-extensions/lib/pipeline/index.js 1.48 kB {1} [built]
 [1752] ./~/websocket-extensions/lib/websocket_extensions.js 4.86 kB {1} [built]
 [1756] multi vendor 400 bytes {1} [built]
     + 2 hidden modules

ERROR in ./~/faye/node/faye-node.js
Module not found: Error: Can't resolve 'fs' in '/Users/johanens/Desktop/work/graphcool/dashboard/node_modules/faye/node'
 @ ./~/faye/node/faye-node.js 2729:13-26
 @ ./~/smooch/lib/utils/faye.js
 @ ./~/smooch/lib/services/conversation-service.js
 @ ./~/smooch/lib/smooch.js
 @ ./~/smooch/lib/main.js
 @ multi vendor

ERROR in ./~/faye/node/faye-node.js
Module not found: Error: Can't resolve 'net' in '/Users/johanens/Desktop/work/graphcool/dashboard/node_modules/faye/node'
 @ ./~/faye/node/faye-node.js 2732:13-27
 @ ./~/smooch/lib/utils/faye.js
 @ ./~/smooch/lib/services/conversation-service.js
 @ ./~/smooch/lib/smooch.js
 @ ./~/smooch/lib/main.js
 @ multi vendor

[...]

PR #384, opened 2016-08-29
Vendor in blueimp-load-image and add browser directive for Faye

### LINKED ISSUE #307: Dependency to blueimp-load-image breaks builds with Webpack
You recently introduced `blueimp-load-image` which seems to break builds with Webpack. See these error messages:
screen shot 2016-07-01 at 18 21 10

lemieux:
This would fix #307 and #357. I basically just vendor in the latest version of blueimpt-load-image and removed all the AMD parts from the UMD loader in each file. It required additional directives in the webpack config to make it work since webpack offers both AMD and CommonJS loaders. 

I also added a directive to tell webpack to use the correct faye version when building for the browser. It seems like it's a bug with webpack 2 where the browser directive of a subdependency is ignored, but since we could do something on our end to mitigate that problem, well... I did something.

I tested those modifications with a side project of mine that uses webpack 2.

@alavers @mspensieri @dannytranlx @jugarrit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,8 @@ node_modules
 bower_components
 *.log
 tmp
-dist/
-lib/
+/dist/
+/lib/
 grunt-aws.json
 grunt-maxcdn.json
 *.sublime-workspace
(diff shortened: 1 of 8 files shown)


mspensieri:
👍

dannytranlx:
👍

PR #386, opened 2016-08-29
Fix intro height computation

lemieux:
Not sure why a min-height was introduced in https://github.com/smooch/smooch-js/pull/378/commits/c7f2a0ff3450e6434b88227252015ea4ae70b9cf, but it was forcing the intro to always be 158px. That caused problems when no channels are active since it leaves a big empty space.

This fix reduces the min-height to be the absolute min-height needed (to fit at least the app icon). The height will also be compute on mount as before. Again, not sure why this was changed.

@alavers @mspensieri @jugarrit @dannytranlx

### DIFF
--- a/src/stylesheets/conversation.less
+++ b/src/stylesheets/conversation.less
@@ -13,9 +13,9 @@
 
     .sk-intro-section {
         background-color: #F8F9FA;
-        padding: 18px 18px 22px 18px;
+        padding: @intro-padding @intro-padding @intro-padding-bottom @intro-padding;
         border-bottom: solid 1px #E6E6E6;
-        min-height: 90px; // app icon height + intro padding
+        min-height: @intro-icon-size + @intro-padding + @intro-padding-bottom; // app icon height + intro padding
 
         .app-name {
             color: #464646;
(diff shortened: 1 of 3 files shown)


mspensieri:
👍

dannytranlx:
👍

src/stylesheets/conversation.less
```diff
@@ -15,7 +15,7 @@
         background-color: #F8F9FA;
         padding: 18px 18px 22px 18px;
         border-bottom: solid 1px #E6E6E6;
-        min-height: 158px;
+        min-height: 90px; // app icon height + intro padding
```
dannytranlx: Ideally we could use vars, but I don't know if we have those values for this situation!

src/stylesheets/conversation.less
```diff
@@ -13,9 +13,9 @@
 
     .sk-intro-section {
         background-color: #F8F9FA;
-        padding: 18px 18px 22px 18px;
+        padding: @intro-padding @intro-padding @intro-padding-bottom @intro-padding;
         border-bottom: solid 1px #E6E6E6;
-        min-height: 158px;
+        min-height: @intro-icon-size + @intro-padding + @intro-padding-bottom; // app icon height + intro padding
```
dannytranlx: nit: there's no need for comment now since we all know how to read :trollface:

PR #387, opened 2016-08-30
Enable scroll to bottom before adding temp message

lemieux:
The problem was that scroll to bottom wasn't enabled before adding the temporary message, so the conversation would only scroll after receiving the response from the server. This fixes it.

@alavers @mspensieri @dannytranlx @jugarrit


jugarrit:
👍

mspensieri:
👍

PR #391, opened 2016-09-06
Fix dropshadow on messenger button icon

lemieux:
It seems that #375 made the dropshadow disappears, but not really the PR's fault. Seems like the PR made `feFlood` be used correctly on the svg, but by working correctly it means no dropshadow. Before the PR, the attribute was just invalid and React didn't add it which made the `feFlood` tag having no effect. For some reason, the valid `flood-color` attribute is just making the shadow disappear.

This PR removes the `feFlood` and brings back the shadow.

image

@alavers @dannytranlx @jugarrit @mspensieri


mspensieri:
👍

Issue #401, opened 2016-09-28
Smooch chat shouldn't show when printing


malthejorgensen:
On pages using the Smooch JS chat widget, chat from within the window shows up when trying to print: [Printed page with Smooch.pdf](https://github.com/smooch/smooch-js/files/497850/Printed.page.with.Smooch.pdf) – this is with the widget being closed.

It's not a big problem, but right now it means that I've added the following to my own CSS:
```
@media print {
  #sk-holder {
    display: none;
  }
}
```

lemieux:
Hey @malthejorgensen, 

sorry about that, clearly we didn't think about printing!

I'll actually add that bit of CSS to our stylesheet if you don't mind me stealing your code! (also, PR are always appreciated :) )

Thanks!

malthejorgensen:
No problem, you can consider that piece of code yours! – usually I do PRs, I guess I was just a bit too busy on that particular day ;)

PR #406, opened 2016-10-11
Hide messenger when printing

lemieux:
It was an quick and easy improvement for #401.

It will hide it when printing a non-embedded version of it.

@dannytranlx @alavers @mspensieri


alavers:
👍

dannytranlx:
👍

PR #408, opened 2016-10-27
Expose Smooch-Core instance and userId

lemieux:
`getUserId` was a missing part of the messenger interface and was needed in order to use `getCore` properly.

I also exposed `getCore` so people could make API calls directly without having the re-expose every single endpoint throught the messenger interface.

@dannytranlx @jugarrit

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ Smooch.updateUser({
 Returns the userId of the current user.
 
 ```javascript
-Smooch.getUser()
+Smooch.getUserId()
 ```
 
 #### getConversation()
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -241,6 +241,13 @@ Smooch.updateUser({
 });
 ```
 
+#### getUserId()
+Returns the userId of the current user.
+
+```javascript
+Smooch.getUser()
```
mspensieri: `getUserId`?

README.md
```diff
@@ -1,12 +1,12 @@
-# [Smooch Javascript SDK](https://smooch.io)
+# [Smooch Web Messenger](https://smooch.io)
```
alavers: 👌

PR #411, opened 2016-10-31
Force scroll to bottom when last message node is undefined

lemieux:
We had some reports where the web messenger was crashing due to `_lastMessageNode` being undefined when the conversation was updated. This PR adds a guard against that.

@mspensieri @jugarrit @dannytranlx

### DIFF
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -86,7 +86,7 @@ export class ConversationComponent extends Component {
         if (messages.length > 0 && messages[messages.length - 1]._id) {
             this._lastTopMessageId = messages[0]._id;
         }
-        
+
         const top = getTop(this._topMessageNode, node);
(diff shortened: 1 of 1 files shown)


src/js/components/conversation.jsx
```diff
@@ -136,13 +136,18 @@ export class ConversationComponent extends Component {
...
-        
+
     };
 
     componentWillUpdate(nextProps) {
         const {messages: currentMessages, isFetchingMoreMessages} = this.props;
         const {messages: newMessages} = nextProps;
 
+        if (!this._lastMessageNode) {
+            this._forceScrollToBottom = true;
```
dannytranlx: Why do we need to scroll to bottom when undefined? Why not leave the conversation view as-is?

lemieux: Because scrolling to bottom is the default behavior to have. When this happens, the conversation is empty and is updating to a non-empty state.

PR #414, opened 2016-11-01
Add Viber channel page

dannytranlx:
image

@hebime @lemieux @jugarrit @mspensieri @wmora


src/js/components/channels/viber-channel-content.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { Component, PropTypes } from 'react';
+import { LoadingComponent } from '../../components/loading';
+import { fetchViberQRCode } from '../../services/integrations-service';
+
+
+export class ViberChannelContent extends Component {
+    static contextTypes = {
```
lemieux: can you convert this to not use `context` at all and use `connect` instead? plsplspls

I'm currently working on converting everything to `connect` and stop using `context`.

src/js/components/channels/viber-channel-content.jsx
```diff
@@ -0,0 +1,40 @@
...
+        ui: PropTypes.object
+    };
+
+    static propTypes = {
+        channelState: PropTypes.object.isRequired
+    };
+
+    render() {
+        const {channelState = {}} = this.props;
```
lemieux: should be in `defaultProps` if you expect default values

dannytranlx: That prop is required anyway, there's no need have a default. I've added the propTypes and remove the defaultProps.

PR #415, opened 2016-11-01
Add Viber channel page

dannytranlx:
See #414.

Merged by mistake, wait until we deploy new backend endpoints.


dannytranlx:
[[Viber] User linking in web messenger](https://trello.com/c/qHid78Eo/4673-viber-user-linking-in-web-messenger)

PR #417, opened 2016-11-02
Replace `context` with `connect`

lemieux:
So... long story short, using React context feature was a mistake. Turns out, there are some cases where updating the context at the top component won't propagate the changes to children (and I assumed it was the case when I chose to use the context over react-redux's connect). 

Why is it a problem? I basically was crashing the embedded mode. Everything was fine in non-embedded mode since everything in the context was dispatch before rendering the messenger, but in embedded mode, you will usually render it before and init after. That was a problem for the introduction component that relied on the context to get the app name and icon. When init is done, the context wasn't getting updated so it was stuck with no app name and no icon.

So this PR fixes it and this is the point where we no longer use the context in this project. 

Most of the changes are related to the tests. Sorry for the pile of changes :(

@dannytranlx @jugarrit @alavers


lemieux:
I think it's 110% more readable and easier to know what to expect from the props... and yeah it was a big one... damn you fat test suite!

src/js/components/introduction.jsx
```diff
@@ -17,12 +17,6 @@ export class IntroductionComponent extends Component {
         appState: PropTypes.object.isRequired
```
jugarrit: Missing some props?

src/js/components/settings.jsx
```diff
@@ -1,24 +1,28 @@
...
 import { hasChannels } from '../utils/app';
 
-export class Settings extends Component {
-    static contextTypes = {
-        settings: PropTypes.object.isRequired
-    };
+export class SettingsComponent extends Component {
 
     static propTypes = {
         className: PropTypes.string
```
jugarrit: Here too?

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -1,27 +1,31 @@
...
+        linkColor: PropTypes.string,
+        phoneNumber: PropTypes.string.isRequired,
+        linkState: PropTypes.oneOf(['unlinked', 'pending', 'linked']),
+        smoochId: PropTypes.string.isRequired,
+        text: PropTypes.object.isRequired,
+        channelState: PropTypes.object.isRequired
     };
 
     linkTwilioNumber = () => {
-        linkTwilioChannel(this.props.userId, {
```
jugarrit: How was this working before?

src/js/components/channels/twilio-channel-content.jsx
```diff
@@ -140,9 +146,17 @@ export class TwilioChannelContentComponent extends Component {
     }
 }
 
-export const TwilioChannelContent = connect((state) => {
+export const TwilioChannelContent = connect(({app, ui: {text}}) => {
     return {
-        ...state.integrations.twilio,
-        userId: state.user._id
```
jugarrit: https://github.com/smooch/smooch-js/pull/417/files#r86180606
Oh, I guess we were always passing in `smoochId` to `ChannelContent` components?

src/js/components/channels/twilio-channel-content.jsx
jugarrit: https://github.com/smooch/smooch-js/pull/417/files#r86180820

lemieux: what do you mean? the component was getting `userId` from the `connect` part... but all channel page are receiving `smoochId` from the parent.

PR #418, opened 2016-11-02
Add isMobile deep link for LINE channel

jugarrit:


src/js/constants/channels.js
```diff
@@ -62,7 +62,11 @@ export const CHANNEL_DETAILS = {
         descriptionKey: 'lineChannelDescription',
         isLinkable: false,
         ...integrationsAssets.line,
-        Component: LineChannelContent
+        Component: !isMobile.phone ? LineChannelContent : undefined,
```
dannytranlx: Are tablets considered phones? Should we display deep links for them too?

jugarrit: Maybe I'll change to any then

src/js/constants/channels.js
```diff
@@ -62,7 +62,11 @@ export const CHANNEL_DETAILS = {
         descriptionKey: 'lineChannelDescription',
         isLinkable: false,
         ...integrationsAssets.line,
-        Component: LineChannelContent
+        Component: !isMobile.any ? LineChannelContent : undefined,
+        getURL: (appUser, {qrCodeUrl}) => {
+            const channelId = /^https?:\/\/qr-official.line.me\/sid\/L\/([a-zA-Z0-9]{3,})\.png$/.exec(qrCodeUrl)[1];
```
jugarrit: After speaking with @lemieux I agreed it would be best to extract this ID and store it serverside to make this transparent for the Web messenger

PR #427, opened 2016-11-08
More lenient Babel versions in peerDeps

Leeds-eBooks:
As babel is concerned only with standards-based transpilation, I think it's safe to assume that any version `6.x.x` of the babel packages will work fine. This change improves DX as it allows us to use the latest version of these packages with whatever bugfixes they include rather than being tied to now-outdated versions of these packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,8 +144,8 @@
     "webpack-dev-server": "1.14.0"
   },
   "peerDependencies": {
-    "babel-polyfill": "6.9.x",
-    "babel-runtime": "6.9.x",
+    "babel-polyfill": "6.x.x",
+    "babel-runtime": "6.x.x",
     "react": "15.x.x",
     "react-dom": "15.x.x"
   },
--- a/package.json
+++ b/package.json
@@ -144,8 +144,8 @@
     "webpack-dev-server": "1.14.0"
   },
   "peerDependencies": {
-    "babel-polyfill": "6.x.x",
-    "babel-runtime": "6.x.x",
+    "babel-polyfill": ">=6.9.0 <7.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -144,8 +144,8 @@
     "webpack-dev-server": "1.14.0"
   },
   "peerDependencies": {
-    "babel-polyfill": "6.9.x",
-    "babel-runtime": "6.9.x",
+    "babel-polyfill": "6.x.x",
```
lemieux: I'd be more comfortable allowing something like `>=6.9.0 <7.0` since we know for sure 6.9.x works and we want to allow new versions.

Makes sense?

Leeds-eBooks: Yep totally fine, I'll update!

Leeds-eBooks:
@lemieux Not 100% sure how this new "request changes" feature works but I think I've done the necessary!

lemieux:
Yeah sorry, I didn't have time to get back to the PR after your changes ;) All good!

lemieux:
Thanks a lot for this :)

Leeds-eBooks:
Great stuff, thanks! How does your release schedule work – will this trigger a release, or will it have to wait for other features/fixes?

lemieux:
We don't have a rigid release schedule for the web messenger. I'll get the two other opened PR merged and cut a release today. I'll keep you in the loop when it's done.

lemieux:
smooch@3.6.1 is out :)

Leeds-eBooks:
Brilliant, thanks so much for the quick turnaround. It's our first time using Smooch and this fills us with confidence!

PR #431, opened 2016-11-09
Add support for quick replies

lemieux:
Soooo, quick replies!

Here's basically how this works : 
- When a message is dispatched to the store, the reducer will extract the quick replies (if present) from it and store them in the store's state. If the message only has quick replies without any text, it's discarded and only the replies are kept.
- From that list of replies, the conversation component figures out if it should show the QuickReplies component or not
- Only the last set of quick replies is shown to the user (hence why I extract them when receiving the message)
- Actions of type `reply` can't be mixed with other action type, so don't worry about it.

@mspensieri @jugarrit @dannytranlx @hebime

### DIFF
--- a/src/js/components/action.jsx
+++ b/src/js/components/action.jsx
@@ -165,7 +165,7 @@ export class ActionComponent extends Component {
                            { buttonText }
                        </button>
                    </div>;
-        } else {
+        } else if (type === 'link') {
             const isJavascript = uri.startsWith('javascript:');
 
             return <div className='sk-action'>
@@ -176,6 +176,8 @@ export class ActionComponent extends Component {
                            { text }
                        </a>
                    </div>;
+        } else {
+            return null;
         }
     }
 }
--- a/src/js/components/action.jsx
+++ b/src/js/components/action.jsx
@@ -165,7 +165,7 @@ export class ActionComponent extends Component {
                            { buttonText }
                        </button>
                    </div>;
-        } else if (type === 'link') {
+        } else if (type === 'link' || type === 'buy' && !stripeIntegration) {
             const isJavascript = uri.startsWith('javascript:');
 
             return <div className='sk-action'>
--- a/src/js/components/action.jsx
+++ b/src/js/components/action.jsx
@@ -17,7 +17,7 @@ export class ActionComponent extends Component {
         text: PropTypes.string.isRequired,
         type: PropTypes.string,
         buttonColor: PropTypes.string,
-        amount: PropTypes.string,
+        amount: PropTypes.number,
         currency: PropTypes.string,
         uri: PropTypes.string,
         state: PropTypes.string,
--- a/src/js/components/action.jsx
+++ b/src/js/components/action.jsx
@@ -165,7 +165,7 @@ export class ActionComponent extends Component {
                            { buttonText }
                        </button>
                    </div>;
-        } else if (type === 'link' || type === 'buy' && !stripeIntegration) {
+        } else if (type === 'link' || (type === 'buy' && !stripeIntegration)) {
             const isJavascript = uri.startsWith('javascript:');
 
             return <div className='sk-action'>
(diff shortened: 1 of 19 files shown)


src/js/components/action.jsx
```diff
@@ -165,7 +165,7 @@ export class ActionComponent extends Component {
                            { buttonText }
                        </button>
                    </div>;
-        } else {
+        } else if (type === 'link' || type === 'buy' && !stripeIntegration) {
```
mspensieri: maybe wrap the && part in parenthesis to reduce ambiguity?

`if (type === 'link' || (type === 'buy' && !stripeIntegration)) {`

src/js/components/conversation.jsx
```diff
@@ -94,21 +95,24 @@ export class ConversationComponent extends Component {
     };
 
     scrollToBottom = () => {
-        const {shouldScrollToBottom} = this.props;
-        if(!this._isScrolling && (shouldScrollToBottom || this._forceScrollToBottom)) {
+        const {shouldScrollToBottom, quickReplies} = this.props;
+        if (!this._isScrolling && (shouldScrollToBottom || this._forceScrollToBottom)) {
             this._isScrolling = true;
-            const timeout = setTimeout(() => {
```
mspensieri: I feel like we keep going back on forth on whether or not `setTimeout` is required. What prompted the change this time?

src/js/components/quick-replies.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+    render() {
+        const {choices, accentColor, isAccentColorDark} = this.props;
+
+        const buttonStyle = {};
+
+        if (accentColor) {
+            const rgb = getRGB(`#${accentColor}`);
+            const {h} = rgbToHsl(...rgb);
+            buttonStyle.backgroundColor = isAccentColorDark ? `hsl(${h}, 100%, 95%)` : `hsl(${h}, 100%, 98%)`;
```
mspensieri: Can it happen where the accent color is the perfect blend and our manipulations actually do nothing to it, so the quick reply just shows as a solid button? i.e. the computed background color is identical or near-identical to the foreground color?

src/js/reducers/conversation-reducer.js
```diff
@@ -118,25 +127,28 @@ export function ConversationReducer(state = INITIAL_STATE, action) {
...
-                messages: state.messages
+                ...INITIAL_STATE,
+                ...action.conversation
             };
         case ConversationActions.SET_MESSAGES:
             return {
                 ...state,
-                messages: assignGroups(sortMessages(removeDuplicates(action.messages)))
+                messages: assignGroups(sortMessages(cleanUpMessages(action.messages))),
+                quickReplies: extractQuickReplies(action.messages[action.messages.length - 1])
```
mspensieri: Curious why quick replies need their own field here? Why not look at the last index of `messages` to see if there are quick replies?

src/js/components/conversation.jsx
lemieux: `this` inside the `setTimeout` callback was the wrong one. `this.props.quickReplies` was full when an appUser message would come in when it should be empty.

I tested it and it's working alright without it. I'd say, no `setTimeout` going forward.

src/js/components/action.jsx
```diff
@@ -176,6 +176,8 @@ export class ActionComponent extends Component {
                            { text }
                        </a>
                    </div>;
+        } else {
+            return null;
```
mspensieri: Is this the correct handling of unknown action types finally? :)

src/js/reducers/conversation-reducer.js
lemieux: That's how it was working before. I changed it because we want to not show messages with replies and no text or image. I started by just omitting them at render time. Then, grouping logic was off, because it considered empty messages in there. So I tried to tweak the logic to ignore empty messages aaaaaaaaaaaaand it was a nightmare with so many edge cases. 

The simple solution was to extract the quick replies from the last message, store them there, and discard any empty message so the grouping and rendering logic wouldn't even need to be changed.

src/js/components/action.jsx
lemieux: Yeah at last ;)

PR #446, opened 2016-12-13
Add support for `text+image+actions`

jugarrit:
Consider message `type`
Split message into parts
Display all parts together nicely

### DIFF
--- a/src/js/components/message.jsx
+++ b/src/js/components/message.jsx
@@ -43,8 +43,25 @@ export class MessageComponent extends Component {
     }
 
     render() {
-        const {name, role, mediaUrl, avatarUrl, text, accentColor, firstInGroup, lastInGroup, linkColor} = this.props;
+        const {name, role, avatarUrl, text, accentColor, firstInGroup, lastInGroup, linkColor, type} = this.props;
         const actions = this.props.actions.filter((a) => a.type !== 'reply');
+
+        const hasText = text && text.trim();
+        const hasImage = type === 'image';
+        const isAppUser = role === 'appUser';
+        const hasActions = actions.length > 0;
+
+        const avatarClass = hasImage ? ['sk-msg-avatar', 'sk-msg-avatar-img'] : ['sk-msg-avatar'];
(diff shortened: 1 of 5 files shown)


lemieux:
Screenshot?

src/js/components/message.jsx
```diff
@@ -43,43 +43,49 @@ export class MessageComponent extends Component {
...
-        const message = mediaUrl ?
-            <ImageMessage {...this.props}
-                          actions={ actions } /> :
-            <TextMessage {...this.props}
-                         actions={ actions } />;
-
-        const containerClass = [mediaUrl ? 'sk-msg-image' : 'sk-msg'];
-
-        const hasContent = (text && text.trim()) || (mediaUrl && mediaUrl.trim());
+        const textPart = hasText && <TextMessage {...this.props}
```
lemieux: since `hasActions` and `hasImage` are passed only to apply a class that adds some style, I think we should just drop those two props and pass the style we want.

lemieux: This way, the text component won't need to be aware of the image or actions concepts.

src/js/components/text-message.jsx
```diff
@@ -32,12 +34,10 @@ export class TextMessage extends Component {
                    </span>;
         });
 
-        if (this.props.text.trim() && this.props.actions.length > 0) {
-            text = <span className='has-actions'>{ text }</span>;
+        if (this.props.hasImage || this.props.hasActions) {
```
lemieux: my other comment refers to this, we use a `has-actions` class for images too, makes little sense.. we coul just pass the style we want and forget about images and actions here.

src/js/components/message.jsx
```diff
@@ -43,43 +43,49 @@ export class MessageComponent extends Component {
...
+                                                 hasActions={ hasActions }
+                                                 hasImage={ hasImage } />;
+        const imagePart = hasImage && <ImageMessage {...this.props}
```
lemieux: same thing here, I think we should just pass the necessary styling to the component instead of making them aware of the concept of actions. (I know it was like that before, but I think it was wrong).

jugarrit:
image
image
image

PR #447, opened 2016-12-15
Tweak CSS to match desired design

jugarrit:
image
image
image

### DIFF
--- a/src/js/components/message.jsx
+++ b/src/js/components/message.jsx
@@ -71,7 +71,10 @@ export class MessageComponent extends Component {
                  src={ avatarUrl } />;
 
         const textClasses = ['sk-message-item', 'sk-message-text'];
-        lastItem === 'text' && textClasses.push('sk-last-item');
+
+        if (lastItem === 'text') {
+            textClasses.push('sk-last-item');
+        }
 
         const textPart = hasText && <TextMessage {...this.props}
                                                  className={ textClasses.join(' ') } />;
@@ -114,7 +117,10 @@ export class MessageComponent extends Component {
                          </div>;
 
         const actionListClasses = ['sk-message-item'];
-        lastItem === 'actions' && actionListClasses.push('sk-last-item');
+
+        if (lastItem === 'actions') {
+            actionListClasses.push('sk-last-item');
+        }
 
         return <div className={ 'sk-row ' + (isAppUser ? 'sk-right-row' : 'sk-left-row') }>
                    { !isAppUser && firstInGroup ? fromName : null }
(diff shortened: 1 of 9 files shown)


src/js/components/message.jsx
```diff
@@ -73,15 +70,9 @@ export class MessageComponent extends Component {
...
-        const extraSpacingStyle = {
-            marginBottom: '5px',
-            display: 'inline-block'
-        };
-
         const textPart = hasText && <TextMessage {...this.props}
-                                                 style={ hasActions || hasImage ? extraSpacingStyle : null } />;
-        const imagePart = hasImage && <ImageMessage {...this.props}
-                                                    style={ hasActions ? extraSpacingStyle : null } />;
+                                                 className={ `sk-message-item sk-message-text${lastItem === 'text' ? ' sk-last-item' : ''}` } />;
```
lemieux: I'd prefer to build this class list in an array and `.join` instead of doing it on a long line like that

src/js/components/message.jsx
```diff
@@ -126,9 +117,11 @@ export class MessageComponent extends Component {
                        <div className={ containerClass.join(' ') }
                             style={ style }
                             ref='messageContent'>
-                           { textPart ? textPart : null }
                            { imagePart ? imagePart : null }
-                           { actionList }
+                           { textPart ? textPart : null }
+                           { hasActions ? <div className={ `sk-message-item${lastItem === 'actions' ? ' sk-last-item' : ''}` }>
```
lemieux: same here

src/js/components/message.jsx
```diff
@@ -73,15 +70,9 @@ export class MessageComponent extends Component {
             <img className={ avatarClass.join(' ') }
                  src={ avatarUrl } />;
 
-        const extraSpacingStyle = {
-            marginBottom: '5px',
-            display: 'inline-block'
-        };
-
         const textPart = hasText && <TextMessage {...this.props}
```
lemieux: why is text not getting `sk-message-item sk-message-text-last-item-long-classname` thing?

jugarrit: What? It is?

lemieux: ugh... github diff told me otherwise... checked again and you are right.

src/js/components/message.jsx
```diff
@@ -110,6 +113,9 @@ export class MessageComponent extends Component {
                              { isAppUser ? '' : name }
                          </div>;
 
+        const actionListClasses = ['sk-message-item'];
+        lastItem === 'actions' && actionListClasses.push('sk-last-item');
```
lemieux: use the `if` luke

PR #448, opened 2016-12-16
Fix image rounding

jugarrit:
Oh, and I added release notes.
And I am a time traveller.


mspensieri:
Attached screenshot reveals a potential bug - why is the bottom right corner of the image rounded?

lemieux:
@mspensieri is right... why rounded corner?

jugarrit:
Maybe because there's a bug... I'll fix it.

PR #454, opened 2016-12-22
Fix round corners for single message

lemieux:
Last fix #451 introduce a regression where messages that were not part of a group would be missing their round corners.

image

This fixes it.

image


lemieux:
Yeah but since `first` set the top corner to 2px and `last` set the bottom corner to 2px, we want to apply those rules only if they are not both first and last.

PR #456, opened 2017-01-02
Fix image upload

mspensieri:
Image upload messages were being created without an `image` type, and
were not correctly handled as images by the UI, causing the
`_restyleBubble` method to throw, and prevent the upload from going
through.

After fixing that issue, it revealed a rendering issue when the `text`
property of a message was equal to its `mediaUrl`, so I fixed that too.


CynthiaNC:
Can I upload media to my own oss bucket?

PR #458, opened 2017-01-17
Embed webpack server in dev server

lemieux:
At last, no need to run `npm run webpack` in a separated console anymore!

With these changes, all you need to run is `npm run dev` and that's it. Webpack is now run as a middleware in the express server.

### DIFF
--- a/config/default/config.json
+++ b/config/default/config.json
@@ -5,6 +5,5 @@
     "USER_ID": "",
     "GIVEN_NAME": "DEFAULT_NAME",
     "SURNAME": "DEFAULT_SURNAME",
-    "EMAIL": "DEFAULT@EMAIL.COM",
-    "SERVER_HOST": "localhost:2992"
+    "EMAIL": "DEFAULT@EMAIL.COM"
 }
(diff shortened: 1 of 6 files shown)


config/default/config.json
```diff
@@ -5,6 +5,5 @@
     "USER_ID": "",
     "GIVEN_NAME": "DEFAULT_NAME",
     "SURNAME": "DEFAULT_SURNAME",
-    "EMAIL": "DEFAULT@EMAIL.COM",
-    "SERVER_HOST": "localhost:2992"
```
dannytranlx: Even without this, will the webpack server be accessible from the network? On a mobile device for example? We used to update that value to the local server IP to make it accessible.

lemieux: yup, webpack assets will be available at the same host and port as the dev server

dannytranlx: [ok_hand](http://pix.iemoji.com/images/emoji/apple/ios-9/256/ok-hand-sign.png)

lemieux:
I'll keep this open but won't merge right now. Hot loader seems broken with the middleware for some reason. I'll debug once I have a couple of minutes.

lemieux:
Hot reloader fixed!

PR #459, opened 2017-01-18
Reduce flag size

jugarrit:


mspensieri:
This is actually changing the resolution of the image, won't we lose quality?

jugarrit:
@mspensieri absolutely, yes. The image size was far too large, however...
Kevin provided the shrunk image.

lemieux:
The image file was like 280kb... it was non-sense.

Did you test if was still working? My fear with resizing the image was that the CSS wouldn't be aligned correctly anymore.

lemieux:
This https://github.com/smooch/smooch-js/blob/integration/src/stylesheets/flags.less

jugarrit:
Of course I tested it!
I wanted the code to be easily adjustable, and it seems that it was.

PR #461, opened 2017-01-19
Redux store refactor

lemieux:
This is a ~somewhat~ big refactor in prevision of the upcoming code splitting story. Last time I tried to do code splitting, one thing that was really blocking it was the fact that the store was being required by pretty much anything. This refactor changes the behavior a bit and the store is only required by the entry point and the rest can have access to it and `dispatch` through `connect`.

I also added `redux-batched-actions` to reduce to amount of rerendering the widget has. Note that "thunks" (async action.... those that return a `(dispatch[, getState]) => {}` function) can't be used in batched actions so you might see some actions being accumulated in an array and then batched and some that won't.

Most of the services method were transformed into redux-thunks. I should probably just move everything in `services` into `actions` and be done with it.

Thanks to the extensive test suite, most of the refactoring quirks were caught and the rest was caught when I tested manually. I think I tested each use case manually to make sure it was working like before.


src/js/components/action.jsx
```diff
@@ -37,9 +38,16 @@ export class ActionComponent extends Component {
...
+
+        bindAll(this, [
+            'onPostbackClick',
+            'onStripeToken',
+            'onStripeClick',
+            'onStripeClose'
+        ]);
     }
 
-    onPostbackClick = (e) => {
```
mspensieri: This syntax doesn't work anymore? Why do we need `bindAll` instead?

src/js/components/alternate-channels.jsx
```diff
@@ -1,15 +1,18 @@
...
-export class AlternateChannels extends Component {
+export class AlternateChannelsComponent extends Component {
     static propTypes = {
         items: PropTypes.array.isRequired
     }
 
-    onChannelClick(event) {
-        showChannelPage(event.target.id);
-    }
+    onChannelClick = (e) => {
```
mspensieri: 😮  seems to work here without `bindAll`?

src/js/smooch.jsx
```diff
@@ -132,33 +133,38 @@ export class Smooch {
...
-            store.dispatch(updateText(props.customText));
+            actions.push(updateText(props.customText));
         }
 
         if (props.serviceUrl) {
-            store.dispatch(AppStateActions.setServerURL(props.serviceUrl));
+            actions.push(AppStateActions.setServerURL(props.serviceUrl));
         }
+
+        store.dispatch(batchActions(actions));
```
mspensieri: wow this must be a fairly decent optimization?

src/js/smooch.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { render, unmountComponentAtNode } from 'react-dom';
 import pick from 'lodash.pick';
+import { batchActions } from 'redux-batched-actions';
 
 import { store } from './stores/app-store';
```
mspensieri: So I guess it's okay for this file to import store since it's the main entry point?

src/js/smooch.jsx
lemieux: I didn't measure it but you can assume it was one render per dispatch before.

src/js/smooch.jsx
lemieux: Yeah the idea is that with code splitting, loading the store could be done asynchronously only if a Smooch method needs it or when it needs to render. Entry point should be as light as possible. When I tried this last, I remember that since the store was required in a lot of places, the bundles webpack was generating were not really worth it.

src/js/components/alternate-channels.jsx
lemieux: eh I forgot that one :p

lemieux:
So reason for using `bindAll` is mainly because of the tests.

using the class variables syntax ( `something = () => {}` ) is equivalent to 

```
constructor() {
     this.something = () => {};
}
```

so it's not available on the prototype.

The problem with that is when you want to stub or spy on a class method (like onClick or onChange), you want to spy/stub the class on the prototype before it gets assign as an event handler on the instance. If you try to spy/stub that method on the instance, the original method is already bound as an event handler so you can't listen on it.

mspensieri:
Would it also work to make sure to invoke the methods on the instance at runtime, rather than bind at create time?

```
onClick= { () => this.onClick() }
```

instead of

```
onClick={ this.onClick }
```

? Just seems a bit odd to me to introduce a new dependency for something like this

lemieux:
Hmmm yeah I think what you are suggesting would work. I was about to write the same util myself and I remembered that lodash had one already to do exactly that.

/shrug I can rewrite them all to do that if we don't want to add lodash.bindall

src/js/smooch.jsx
mspensieri: > only if a Smooch method needs it or when it needs to render

Curious in which case you can have the widget load without the store? Just calling `.init()` will trigger you to load the store, right?

lemieux: yup but that's exactly what I mean, the store is not required until you call a method on Smooch. It would lighten the entry point by not loading the store (which pulls in the reducers and the actions).

lemieux: Using webpack code splitting, requiring a module becomes a promise so you just need to chain it.

mspensieri: okok so lazy-loading the store speeds up time to load the initial script, but doesn't really affect time to first meaningful render, right?

lemieux: it could affect the time to first render in theory if we also do a smart job on lazy loading the ui

dannytranlx:
> Thanks to the extensive test suite, most of the refactoring quirks were caught and the rest was caught when I tested manually. I think I tested each use case manually to make sure it was working like before.

😍 🙏

alavers:
👀

src/js/components/chat-input.jsx
```diff
@@ -1,61 +1,71 @@
...
     constructor(...args) {
         super(...args);
-
-        this.state = {
-            text: ''
-        };
-
-        this.onChange = this.onChange.bind(this);
-        this.onSendMessage = this.onSendMessage.bind(this);
+        bindAll(this, [
```
alavers: I think bindAll accepts rest parameters, so you could send separate string args instead of an array. But maybe you prefer on their own lines? If so this is fine.

src/js/smooch.jsx
alavers: also, while you're at it, should this be named `store/app.js` now?

lemieux: better yet... I'll rename that file `./store/index` ... we only have one store..

src/js/services/auth.js
```diff
@@ -0,0 +1,7 @@
+import { core } from './core';
```
alavers: Is this only used in more than one place? Does it deserve to be its own service?

src/js/smooch.jsx
alavers: :doit:

src/js/services/auth.js
lemieux: It could go in the app one I guess.

src/js/utils/dom.js
```diff
@@ -73,16 +72,19 @@ export function stopMonitoringUrlChanges() {
...
-function onWindowFocus() {
-    store.dispatch(hasFocus(true));
-}
+let onWindowFocus;
+let onWindowBlur;
 
-function onWindowBlur() {
-    store.dispatch(hasFocus(false));
-}
+export function monitorBrowserState(dispatch) {
```
alavers: any risk of this being called more than once? do we need some way to remove listeners?

lemieux: right.. I'll check if they already exist.

PR #466, opened 2017-01-24
Fix UI problems

jugarrit:
Resolves a slew of UI issues including:

- Quick replies don't trigger scroll to bottom behaviour
- Bottom margin in every `first-message` was missing
- Word breaks added ugly hyphenation
- Auto-scrolling on receiving an App Maker message was not working in IE


jugarrit:
@mspensieri 
image
image

PR #470, opened 2017-01-27
Replace FB button with m.me link

lemieux:
In prevision of the iframized version of the Web Messenger, one blocker was the FB button. Facebook now supports passing data as referral to a Messenger link and we can leverage that to do the user linking. It also means that : 
 - user linking works even for people with script blockers
 - we won't clash with an existing FB sdk anymore
 - we have one less dependency to load in the page

:tada: 

Of course this PR can't be merged and release before the backend knows how to handle the new events.


lemieux:
@mspensieri There's no screenshot I can show really.. clicking on the Messenger icon brings you directly to it with this (like Telegram). It doesn't bring you to a subpage where you need to click on a button.

I could gif it, but it's pretty much straightforward.

package.json
```diff
@@ -48,7 +48,6 @@
     "react-addons-css-transition-group": "15.1.0",
     "react-document-title": "2.0.1",
     "react-imageloader": "2.1.0",
-    "react-messenger-plugin": "1.0.1",
```
dannytranlx: BibleThump

PR #471, opened 2017-01-31
Add message retry functionality

jugarrit:
message retry
send_state_with_animation

### DIFF
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -218,6 +218,7 @@ export class ConversationComponent extends Component {
                                      accentColor={ accentColor }
                                      linkColor={ linkColor }
                                      onLoad={ this.scrollToBottom }
+                                     entity={ message }
                                      {...message}
                                      lastInGroup={ lastInGroup } />;
         });
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -218,7 +218,6 @@ export class ConversationComponent extends Component {
                                      accentColor={ accentColor }
                                      linkColor={ linkColor }
                                      onLoad={ this.scrollToBottom }
-                                     entity={ message }
                                      {...message}
                                      lastInGroup={ lastInGroup } />;
         });
--- a/src/js/components/message.jsx
+++ b/src/js/components/message.jsx
@@ -8,6 +8,7 @@ import { Action } from './action';
 import { findDOMNode } from 'react-dom';
 import { getElementProperties } from '../utils/dom';
 import { resendMessage } from '../services/conversation';
+import { SEND_STATUS } from '../constants/message';
 
 class Message extends Component {
     static propTypes = {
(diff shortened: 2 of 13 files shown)


src/js/components/conversation.jsx
```diff
@@ -218,6 +218,7 @@ export class ConversationComponent extends Component {
                                      accentColor={ accentColor }
                                      linkColor={ linkColor }
                                      onLoad={ this.scrollToBottom }
+                                     entity={ message }
                                      {...message}
```
lemieux: if you are going to pass the whole message as a prop, remove the spreading here and use props from the entity directly instead, it will be much cleaner instead of having the same data at two places.

src/js/components/message.jsx
```diff
@@ -16,11 +19,14 @@ export class MessageComponent extends Component {
...
-        lastInGroup: PropTypes.bool
+        lastInGroup: PropTypes.bool,
+        sendStatus: PropTypes.string,
+        entity: PropTypes.object.isRequired
     };
 
     static defaultProps = {
-        actions: []
+        actions: [],
+        sendStatus: 'sent'
```
lemieux: can you make a `SEND_STATUS` constant somewhere and have the status string there?

src/js/reducers/conversation-reducer.js
```diff
@@ -10,21 +10,10 @@ const INITIAL_STATE = {
 };
 
 const sortMessages = (messages) => messages.sort((messageA, messageB) => {
-    // received is undefined when it's the temp message from the user
-    if (!messageA.received && !messageB.received) {
```
lemieux: much cleaner!

jugarrit:
send_state_with_animation

src/js/services/conversation.js
```diff
@@ -15,6 +15,59 @@ import { hasLinkableChannels, getLinkableChannels, isChannelLinked } from '../ut
 import { CONNECT_NOTIFICATION_DELAY_IN_SECONDS } from '../constants/notifications';
 import { getUserId } from './user';
 
+const postSendMessage = (dispatch, getState, message) => {
```
lemieux: these functions are actions, they shouldn't be here. 

You can certainly move them in the conversation actions and transform them as 
```
    return (dispatch, getState) => { ... };
};
```
instead.

lemieux: Then you do `dispatch(postSendMessage(message))` whererever you need to execute.

src/js/services/conversation.js
```diff
@@ -15,6 +15,59 @@ import { hasLinkableChannels, getLinkableChannels, isChannelLinked } from '../ut
...
+    actions.push(replaceMessage({
+        _clientId: message._clientId
+    }, response.message));
+
+    dispatch(batchActions(actions));
+    observable.trigger('message:sent', response.message);
+    return response;
+};
+
+const onMesageSendFailure = (dispatch, getState, message) => {
```
lemieux: missing an `s`

PR #477, opened 2017-02-03
Add location support

jugarrit:
location_failed
location_retry
location


jugarrit:
location_failed
location_retry
location

src/js/services/conversation.js
```diff
@@ -66,99 +67,69 @@ const onMessageSendSuccess = (message, response) => {
...
-                }
-            }
+            return newMessage;
         }
 
-        return response;
-    };
-}
-
-export function sendChain(sendFn) {
```
mspensieri: Gonna be some painful merge conflicts in this branch lol

lemieux: why do you have something else going on?

jugarrit: Perhaps :P

mspensieri: https://github.com/smooch/smooch-js/pull/483

lemieux:
@mspensieri it is, in the first gif you can see him hovering on top of it and the underline appears. not sure why we dropped the underline without hover though.

src/js/actions/conversation-actions.js
```diff
@@ -32,9 +32,7 @@ export function setMessages(messages) {
 export function addMessage(props) {
     return {
         type: ADD_MESSAGE,
-        message: Object.assign({
-            actions: []
-        }, props)
+        message: Object.assign({}, props)
```
lemieux: `{...props}`

src/js/constants/message.js
```diff
@@ -3,3 +3,9 @@ export const SEND_STATUS = {
     FAILED: 'failed',
     SENT: 'sent'
 };
+
+export const LOCATION_ERRORS = {
```
mspensieri: I only see `PERMISSION_DENIED` used, are these codes generated by the browser or are we supposed to be using these values ourselves?

src/js/components/reply-actions.jsx
```diff
@@ -0,0 +1,97 @@
...
+        dispatch(sendMessage(text));
+    }
+
+    onSendLocation() {
+        const {dispatch, locationNotSupportedText} = this.props;
+
+        if ('geolocation' in navigator) {
+            dispatch(sendLocation());
+        } else {
+            alert(locationNotSupportedText);
```
mspensieri: Is this a literal alert (like `window.alert`) or is it something else?

src/js/components/loading.jsx
```diff
@@ -3,24 +3,68 @@ import React, { Component } from 'react';
...
-                <div className='sk-circle6 sk-circle'></div>
-                <div className='sk-circle7 sk-circle'></div>
-                <div className='sk-circle8 sk-circle'></div>
-                <div className='sk-circle9 sk-circle'></div>
-                <div className='sk-circle10 sk-circle'></div>
-                <div className='sk-circle11 sk-circle'></div>
-                <div className='sk-circle12 sk-circle'></div>
+            <div style={ style }
+                 className={ classNames.join(' ') }>
+                <div className='sk-circle1 sk-circle'>
```
lemieux: could we just replace this with something like 
```
const circles = [];
for(let i = 1; i <=12; i++) {
    circles.push(<div key={i} className={`sk-circle${i} sk-circle`>
                     <div className='sk-inner-circle'
                          style={ innerCircleStyle } />
                 </div>);
}


// then use it
{circles}
```

src/js/components/reply-actions.jsx
lemieux: If the browser doesn't support sending location, I'm not sure the button should appear at all.

src/js/constants/message.js
lemieux: I'm fine with mapping the browser errors even if we don't use them all.

src/js/services/conversation.js
lemieux: ah lol

test/specs/components/reply-actions.spec.js
```diff
@@ -21,23 +36,19 @@ describe('QuickReplies Component', () => {
...
-                    web: {
-                        accentColor: '',
-                        isAccentColorDark: true
-                    }
-                }
-            }
-        });
         sandbox.stub(conversationService, 'sendMessage');
+        sandbox.stub(conversationService, 'sendLocation');
+        navigator.geolocation = true;
```
lemieux: We should keep the default value and reassign it in `afterEach`. It could have unwanted side effects on future tests.

src/js/components/reply-actions.jsx
jugarrit: It's a literal alert.
Talked about this with JP, there are very few browsers that don't support it.
If the button doesn't appear, what do we put in its place? We could use the fallback text, but it introduces troublesome workaround code.

src/js/constants/message.js
jugarrit: They're generated by the browser, I figured they were nice to have even if unused.

src/js/components/reply-actions.jsx
lemieux: Maybe rendering an unclickable button with the error text in it?

lemieux: Browser support is broad enough, but we might have problem with people who just disabled the geolocation (and it might not be their fault.. corporate policies or something like that).

src/js/components/message.jsx
```diff
@@ -57,14 +62,14 @@ class Message extends Component {
...
-        const actions = this.props.actions.filter((a) => a.type !== 'reply');
+        const {name, role, avatarUrl, text, accentColor, firstInGroup, lastInGroup, linkColor, type, mediaUrl, sendStatus, clickToRetryText, tapToRetryText, locationSendingFailedText} = this.props;
+        const actions = this.props.actions.filter(({type}) => type !== 'reply' && type !== 'locationRequest');
         const hasText = text && text.trim() && text.trim() !== mediaUrl;
         const hasImage = type === 'image';
+        const hasLocation = type === 'location';
         const isAppUser = role === 'appUser';
         const hasActions = actions.length > 0;
-        const lastItem = hasActions ? 'actions' : hasText ? 'text' : null;
+        const lastItem = hasActions ? 'actions' : hasText ? 'text' : hasLocation ? 'location' : null;
```
jugarrit: This is for ones that are currently uploading.
If it has text, we are displaying that text as normal, and so the lastItem displayed will in fact be `text`.

src/js/components/reply-actions.jsx
jugarrit: After talking with @jpjoyal we decided that we want to have the same experience on all devices, and currently we're displaying a dialog when there is an error of this sort, so we will keep this functionality.

README.md
```diff
@@ -155,8 +155,20 @@ var skPromise = Smooch.init({
...
+        smsContinue: 'Continue',
+        smsCancel: 'Cancel',
+        fetchingHistory: 'Retrieving history...',
+        fetchHistory: 'Load more',
         clickToRetry: 'Message not delivered. Click to retry.',
-        tapToRetry: 'Message not delivered. Tap to retry.'
+        tapToRetry: 'Message not delivered. Tap to retry.',
+        locationServicesDenied: 'Access to location services was denied. To enable, go to your browser\'s privacy settings and grant access.',
+        locationSendingFailed: 'Could not send location',
+        locationNotSupported: 'Your browser does not support location services. Please send your location via text.'
```
jpjoyal: Suggest: "Your browser does not support or have location services disabled. Please type your address."

src/js/services/conversation.js
```diff
@@ -163,12 +144,58 @@ export function resendMessage(messageClientId) {
...
+            }
         }
 
         return dispatch(sendChain(postUploadImage, newMessage));
     };
 }
 
+export function sendLocation(message) {
+    return (dispatch, getState) => {
+        const locationMessage = message || dispatch(addMessage(null, {
```
lemieux: I don't know how I feel about passing `null` for the text with extra props when we could just to `addMessage({text: 'something', type: 'location' })` there the object would just merge with the set of defaults in `addMessage`. You could then just omit `text` in there and it would be much cleaner.

src/js/services/conversation.js
```diff
@@ -177,18 +204,11 @@ export function uploadImage(file) {
 
         return resizeImage(file)
             .then((dataUrl) => {
-                const message = {
+                const message = dispatch(addMessage(null, {
```
lemieux: yeah.. I don't like the concept of `extra` props :(

PR #483, opened 2017-02-06
Fix issue where client would connect to faye before conversation started

jugarrit:


src/js/services/conversation.js
```diff
@@ -113,84 +105,67 @@ export function handleConnectNotification(response) {
...
+        const postSendHandler = (response) => {
+            return Promise.resolve(dispatch(onMessageSendSuccess(message, response)))
+                .then(() => dispatch(setShouldScrollToBottom(true)))
+                .then(() => dispatch(handleConnectNotification(response)))
                 .then(() => dispatch(connectFayeConversation()))
-                .then(() => sendFn())
-                .then(enableScrollToBottom)
-                .then((response) => dispatch(handleConnectNotification(response)));
-        }
+                .catch();
```
lemieux: Please add a comment about why this swallows errors,

src/js/services/conversation.js
```diff
@@ -199,27 +174,23 @@ export function uploadImage(file) {
 
         if (!isFileTypeSupported(file.type)) {
             dispatch(showErrorNotification(getState().ui.text.invalidFileError));
-            return Promise.reject('Invalid file type');
```
mspensieri: Why remove this rejection? Should it at least `return`?

jugarrit: Yes, it should definitely return.
I removed the rejection for the same reason we have a catchall in `sendChain`. We're handling the `exception` here and there's no need to bubble it up to the UI

PR #484, opened 2017-02-07
Track current location in store and update button icon with location

lemieux:
Some people reported problems where the icon in the messenger button wasn't showing up on Firefox once again. Last time we fixed that by using the full url for the drop shadow filter url but we we're listening to url changes. This PR now tracks location changes and store the new location in the store so any component can be notified and updated when it changes.

I also changed the filter `id` for some random value to prevent clashing with anything with the id `dropShadow` in the host page.


jugarrit:
Wondering if `location` is confusing... We're using `currentUrl` when we update the device in `updateNowViewing`. Could we use `currentUrl` instead?

lemieux:
We need the whole location object so we can pick the parts that we want instead of using currentUrl and try to parse and decompose it.

PR #486, opened 2017-02-10
Allow calling of `sendMessage` with props OR text

jugarrit:


test/specs/services/conversation.spec.js
```diff
@@ -240,7 +240,20 @@ describe('Conversation service', () => {
 
         conversationStartedSuite(conversationService.sendMessage('message'));
 
-        it('should add message and send message', () => {
+        it('should add message and send message', function*() {
+            mockedStore = createMockedStore(sandbox, getProps());
+
+            yield mockedStore.dispatch(conversationService.sendMessage('message'));
```
lemieux: should test the outcome of this though

test/specs/services/conversation.spec.js
```diff
@@ -240,7 +240,20 @@ describe('Conversation service', () => {
...
+            yield mockedStore.dispatch(conversationService.sendMessage({
+                type: 'text',
+                'text': 'message'
```
lemieux: no need for `' '` around `text` here

test/specs/services/conversation.spec.js
jugarrit: We are? We're checking that `postSendMessage` is called - one for each message.
Unless you mean check the args, too? I can do that.

lemieux: Yeah I mean, does it transfer that string arg to message.text properly.

src/js/services/conversation.js
```diff
@@ -91,12 +91,13 @@ const addMessage = (text, extra = {}) => {
...
-            ...extra
+            sendStatus: SEND_STATUS.SENDING
         };
 
-        if (text) {
-            message.text = text;
+        if (typeof props === 'string') {
+            message.text = props;
+        } else {
+            Object.assign(message, ...props);
```
mspensieri: How is the spread working here? Shouldn't this be either

```
Object.assign(message, props)
```

or

```
message = {
   ...message,
   ...props
}
```

?

jugarrit: I don't know why it's working @mspensieri but I'll change it so that it makes more sense to us

lemieux: I guess it spreads the object as multiple objects with only one key and one value and `Object.assign` accepts multiple objects in a row

PR #491, opened 2017-02-16
Fix myriad of small bugs

jugarrit:


src/js/services/conversation.js
```diff
@@ -175,9 +175,23 @@ export function sendLocation(props = {}) {
...
+            let timedOut = false;
+
+            const timeout = setTimeout(() => {
+                timedOut = true;
+                dispatch(onMessageSendFailure(message));
+                resolve();
+            }, 10000);
+
             navigator.geolocation.getCurrentPosition((position) => {
+                clearTimeout(timeout);
```
lemieux: shouldn't it be inside the `if(timeout)` ?

lemieux: ah nvm I read that wrong

src/js/services/conversation.js
```diff
@@ -192,9 +206,16 @@ export function sendLocation(props = {}) {
...
                     .then(resolve);
             }, (err) => {
+                clearTimeout(timeout);
+                if (timedOut) {
+                    return;
+                }
                 if (err.code === LOCATION_ERRORS.PERMISSION_DENIED) {
                     setTimeout(() => alert(locationServicesDeniedText), 100);
                     dispatch(removeMessage(message._clientId));
+                } else if (err.code === LOCATION_ERRORS.POSITION_UNAVAILABLE && location.protocol !== 'https') {
```
jugarrit: Chrome uses `LOCATION_ERRORS.PERMISSION_DENIED` for the `https:` problem. Will change this case to only check protocol.

PR #495, opened 2017-02-19
unreadCount event

fsalomon:
Usage:
```
Smooch.on('unreadCount', function(unreadCount) {
    console.log('the number of unread messages was updated', unreadCount);
});
```

### DIFF
--- a/src/js/components/message-indicator.jsx
+++ b/src/js/components/message-indicator.jsx
@@ -1,6 +1,5 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
-import { observable } from '../utils/events';
 import DocumentTitle from 'react-document-title';
 
 const BLINKING_INTERVAL = 1500;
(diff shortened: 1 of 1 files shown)


lemieux:
It will also need an addition to this section in the readme https://github.com/smooch/smooch-js/blob/integration/README.md#events

src/js/components/message-indicator.jsx
```diff
@@ -70,6 +71,7 @@ export class MessageIndicatorComponent extends Component {
     }
 
     componentWillReceiveProps({unreadCount}) {
+        observable.trigger('unreadCount', unreadCount);
```
lemieux: This should be triggered here https://github.com/smooch/smooch-js/blob/integration/src/js/smooch.jsx#L102 instead right after `handleNotificationSound();`.

fsalomon:
Thanks! Changes done.

lemieux:
Just to let you know it went out yesterday with the 3.13.0 release. Thanks again for your contribution!

PR #498, opened 2017-02-27
unreadCount event

fsalomon:
The event should be triggered outside the `if (unreadCount > 0) { }`. Otherwise it defeats most of its purpose.

### DIFF
--- a/src/js/smooch.jsx
+++ b/src/js/smooch.jsx
@@ -100,10 +100,10 @@ function onStoreChange({messages, unreadCount}) {
                     initialStoreChange = false;
                 } else {
                     handleNotificationSound();
-                    observable.trigger('unreadCount', unreadCount);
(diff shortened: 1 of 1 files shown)


alavers:
Thanks for the PR! The only issue with this fix is that it will also trigger an unreadCount=0 event for every message the user sends. I've filed an issue on our end to design a deeper fix for this problem.

alavers:
I'm just realizing now you were the author of this code in the first place hah! I suppose you're OK with unreadCount=0 events being fired on every outgoing user message?

fsalomon:
Yes I am :-)

I don't mind multiple events being fired. As long as it won't freeze the UI ;-)

alavers:
Fair enough!

alavers:
It's live

PR #501, opened 2017-03-05
fix React JSX Error for 'for' props in label tag

ali-master:
Replace for with htmlFor in label props

https://github.com/facebook/react/issues/1819


lemieux:
Thanks for this! That went unnoticed :confused: .

That should go out with the next release.

PR #502, opened 2017-03-06
Use onScroll instead of onWheel

lemieux:
Using onWheel was causing significant performance issue because whenever onWheel is used, React binds an event on the page directly and not just the target. We were using onWheel to reset the unread count but only when the user is scroll with their mousewheel. I switch it to use onScroll instead. The downside is that the event is trigger also when the conversation is scrolled down with a new message, but the operation is a noop when unread count is 0.


lemieux:
After discussions, using onScroll would  trigger other side effects. We decided to just stop using it. We have other triggers to reset the unread count.

PR #507, opened 2017-03-22
Link channels via transferrequest API

alavers:
Applies to messenger, telegram and viber

### DIFF
--- a/src/js/services/app.js
+++ b/src/js/services/app.js
@@ -1,6 +1,7 @@
 import * as AppStateActions from '../actions/app-state-actions';
 import { preventMobilePageScroll, allowMobilePageScroll } from '../utils/dom';
 import { resetUnreadCount, connectFayeUser } from './conversation';
+import { fetchTransferRequestCode } from './integrations';
 import { observable } from '../utils/events';
 import { hasLinkableChannels, isChannelLinked } from '../utils/user';
 import { getIntegration } from '../utils/app';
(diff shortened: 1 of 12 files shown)


alavers:
Adding some tests...

src/js/services/app.js
```diff
@@ -74,24 +75,41 @@ export function showChannelPage(channelType) {
...
+            dispatch(AppStateActions.showChannelPage(channelType));
+
+            return dispatch(connectToFayeUser())
+                .then(() => dispatch(channelDetails.onChannelPage()));
+        }
+
+        function openLink() {
             const appChannel = getIntegration(integrations, channelType);
-            const link = channelDetails.getURL(user, appChannel, isLinked);
+            const getCode = channelDetails.usesCode ?
```
lemieux: What's the behavior here? I'm surprised it doesn't trigger the browser popup blocker with all of this since we are calling `window.open` so late in the process. Otherwise, do we have some sort of ui feedback while it waits for a code?

alavers: I didn't implement any UI feedback, it didn't seem necessary since it's a pretty fast API.

Wasn't aware of the popup blocker danger. Assuming the code was safe before, do you know how popup blockers were linking the UI event to the window.open before this change? Do you know much of a delay might cause popup blocker to intervene?

alavers:
Added a pretty low hanging fruit unit test. Doesn't seem like `app.showChannelPage` is very well covered, am I missing something somewhere?

src/js/services/app.js
lemieux: the popup blocker isn't triggered if `window.open` is called in a direct response to a user action... I think once you add a promise in the mix, you exit the current event stack and the browser doesn't consider it in response to a user action and triggers the popup blocker.

alavers:
For context, to address the popup blocker issue I'm going to give messenger and telegram their own config panels like the rest of the channels. This changes a lot in this PR, I'll just open a new one.

*Hi, I'm Kevin Gauthier and I approve this UI change*

PR #508, opened 2017-03-23
Throttle getMessages calls

alavers:
I have the default throttle interval set to 5 seconds, open to feedback on that.

Note that this throttling does not apply to get scrolling pagination, and each user gets their own throttle, so if you log in as a different user you won't have to wait 5 seconds to see your conversation.

### DIFF
--- a/src/js/utils/throttle.js
+++ b/src/js/utils/throttle.js
@@ -1,0 +1,32 @@
+
+/**
+ * Throttle for functions that return a promise. If additional calls are made
+ * while the throttle is on, the original promise will be returned.
+ *
+ * Usage:
+ *
+ * const func = () => Promise.resolve('foo')
+ *
+ * const throttle = new Throttle();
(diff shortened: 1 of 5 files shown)


src/js/utils/throttle.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Usage:
+ *
+ * const func = () => Promise.resolve('foo')
+ *
+ * const throttle = new Throttle();
+ * const throttledFunc = () => throttle.exec(() => func());
+ * throttledFunc().then((foo) => {})
+ */
+export class Throttle {
```
jpjoyal: Surprised we didn't have this in our toolbox via a 3rd party lib

alavers: 3rd party promise throttles do exist, but I feel it's below the complexity threshold to justify adding a dependency.

PR #509, opened 2017-03-27
Link channels via transferrequest API

alavers:
Applies to messenger, telegram and viber

(a second take on https://github.com/smooch/smooch-js/pull/507 that is popup blocker friendly)

messenger
telegram
viber
list

Unit tests coming


alavers:
Note also that Viber has a channel page on both desktop and mobile. It's a QR code on desktop, and a `viber://` link on mobile. (the above gif shows what it looks like in mobile)

src/js/components/channels/transfer-request-channel-content.jsx
```diff
@@ -0,0 +1,62 @@
...
+export class TransferRequestChannelContentComponent extends Component {
+
+    static propTypes = {
+        channelState: PropTypes.object.isRequired,
+        transferError: PropTypes.string
+    };
+
+    onLink = () => {
+        const {dispatch, type} = this.props;
+        dispatch(resetTransferRequestCode(type));
```
lemieux: if those actions aren't async (using thunks), you can batch them using `batchActions`. It will optimize rendering.

src/js/components/channels/transfer-request-channel-content.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+        if (hasError) {
+            return <a className={ 'sk-error-link' }
+                      onClick={ this.onTryAgain }>
+                      { transferError }
+                   </a>;
+        }
+
+        if (transferRequestCode) {
+            return <a alt='Connect'
```
lemieux: Should this be styled as a button instead of a small link?

src/js/components/channels/transfer-request-channel-content.jsx
```diff
@@ -0,0 +1,62 @@
...
+                      target='_blank'
+                      onClick={ this.onLink }
+                      href={ url }>Connect</a>;
+        }
+
+        const loadingStyle = {
+            height: 40,
```
lemieux: use units pls! It's already on its way out to not use units in React.

src/js/components/channels/transfer-request-channel-content.jsx
alavers: I'll @kvgauthier ize it

PR #528, opened 2017-04-13
Call unthrottled getMessage action when connecting faye

### LINKED ISSUE #527: Smooch.sendMessage no longer gets a reply
I'm using an embedded widget, and using `Smooch.sendMessage()` when the widget is ready.
The message is sent, but I don't get a reply. 
Here's a minimal test-case: https://jsfiddle.net/49jwet9r/5/

After `sendMessage("Welcome")` a "Pleased to meet you" whisper should be received. 
The whisper is there if I look at [the logs](https://i.imgur.com/U4gEVCH.png) but nothing comes back on the front-end. If you simply re-run the fiddle you'll see both messages there, as expected.

Since I know this was working for me before, I built smooch.js **v3.13.3** from source, and the problem is not there. 
See the same demo using that older version: https://jsfiddle.net/49jwet9r/8/

Expected behavior.

lemieux:
This should fix #527. In this example, `getConversation` is called before sending a message which triggers the throttling of `getMessage` which then makes the one on `transport:up` meaningless because it gets throttled.

This makes sure we call the unthrottled version of `getMessages` when faye connects no matter what. It's really the only place where throttling shouldn't happen so we don't miss messages.

### DIFF
--- a/src/js/components/conversation.jsx
+++ b/src/js/components/conversation.jsx
@@ -26,7 +26,7 @@ export class ConversationComponent extends Component {
         embedded: PropTypes.bool.isRequired,
         shouldScrollToBottom: PropTypes.bool.isRequired,
         isFetchingMoreMessages: PropTypes.bool.isRequired,
-        hasMoreMessages: PropTypes.bool.isRequired,
+        hasMoreMessages: PropTypes.bool,
         introHeight: PropTypes.number,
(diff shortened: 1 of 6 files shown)


lemieux:
Hmm I'll close this since we added throttling just pretty much to cover excessive `transport:up` on a flaky connection.

PR #531, opened 2017-04-14
upgrade React from 15.1.0 to 15.5.4

ali-master:
Hi there 👋

First things first, thanks for all the work put into this awesome library!

In one of the projects I use it in, we just updated React to 15.5, which brought in some [deprecation warnings](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#new-deprecation-warnings).

Thanks.


lemieux:
Hey @ali-master,

sorry it took so long to reply, we just didn't have time to handle this yet. We took a quick look at this and it's probably something we'll merge in (tests need to be fixed though). 

We also need to make sure everything is still good for people using 15> X <15.5.4 since React is a peer dependency. After some reading, I don't think there will be problems with that, but we still need to make sure.

We're a bit busy at this moment, but we'll make sure to take a deeper look at this when we have a sec.

Thanks for this :)

lemieux:
As we're working on a new feature and we upgraded a lot of dependencies (including React), I'll close this PR since it might create too many conflicts (and we pretty much did the same thing).

Sorry about that and thanks for the PR!

PR #534, opened 2017-05-10
Send invalidation to Cloudfront instead of MaxCDN

lemieux:
We're migrating away from MaxCDN to Cloudfront. This is it.


lemieux:
;) I tested it manually and checked that Cloudfront was receiving the invalidation request.

PR #535, opened 2017-05-15
Set cache-control headers when uploading to s3

lemieux:
This will help leverage browser cache with our static assets. I set the value on smooch.min.js to 5 min so it doesn't take too long for a new deployment to roll out.


nathang21:
Any chance we could up this a bit higher, especially for the image assets or some of the other things that essentially never change? Would make a big difference.

mspensieri:
@nathang21 This release process, and the associated values, are no longer valid in the 4.x version stream. If there is a specific asset that has a low cache expiration time, please open a new issue with the details

PR #538, opened 2017-05-30
First pass at rendering and running inside an iframe

lemieux:
As this feature will require big changes, I'm trying to break it down in multiple PRs. This one is already big enough, sorry!

## So what does this PR covers?
- Splitting logic between host code and frame code
- Full handling of the badge version of the widget (animations, etc.)
- Upgrade to webpack 2
- A bit of a clean up of the files organization

## What doesn't it cover?
- Tests
- Tab and embedded version of the widget
- Build/packaging/deployment process
- npm version of the lib
- Full cleanup of the code base (pretty sure I can remove useless stuff now)
- Style customization (out of scope)

## How does it work
So basically now we have two parts in the library : host and frame.

The host part contains a skeleton of `Smooch` with only `init` and `on/off`on it. Those can be called and the calls are cached until the frame part is loaded. Once the frame part is loaded, the skeleton is hydrated with all the `Smooch` public methods and cached calls are properly applied to the real methods. The methods on the skeleton are the only one that should be called when the lib is not fully init'd.

To load the frame part, the host part creates an iframe and injects the most basic html needed to load the other part. At this point, the frame part is loaded using `<script src="/_assets/frame.js"></script>`, but this method won't work for the npm version but this will be covered in another PR.

### Media queries
Since media queries don't work inside the iframe (the size of the iframe is the "window" size in that context so it's always too small), I included a new dependency (enquire.js) to manage and detect media queries in the host page and relay that information to the frame. The widget in the iframe will then apply to itself a class (widget-xs, widget-sm, widget-md, widget-lg) so that it knows how to render itself.

### Notes
- You might be wondering why the code in the host lib looks so old and not too much es2015... well.. I tried as much as I can to not use the new stuff (Array.each, etc.) as those are new things that older browser don't support and the host lib doesn't have polyfills (on purpose! the frame has them though).
- This has been tested on Chrome, Android, iOS and IE10 so far.

It think this cover the high-level stuff, but surely more questions will come from the review.


src/stylesheets/reset.less
```diff
@@ -1,28 +0,0 @@
-/* Eric Meyer's Reset CSS v2.0 - http://cssreset.com */
```
lemieux: I opted to use the bootstrap one since it's a bit more aggressive and we're now contained inside an iframe.

src/host/js/smooch.js
```diff
@@ -0,0 +1,93 @@
...
+    iframe = document.createElement('iframe');
+    iframe.className = hostStyles.ref().locals.iframe;
+    let loaded = false;
+    iframe.onload = () => {
+        if (!loaded) {
+            loaded = true;
+            delete iframe.onload;
+            const doc = iframe.contentWindow.document;
+            doc.open();
+            doc.write('<!DOCTYPE html><html><head><script src="/_assets/frame.js"></script></head></html>');
```
lemieux: this will change as part of a future PR about packaging the lib.

make-webpack-config.js
```diff
@@ -19,41 +19,80 @@ module.exports = function(options) {
...
+        assets: './src/frame/js/constants/assets'
     } : {
-        smooch: ['./src/js/utils/polyfills', './src/js/umd']
+        host: ['./src/host/js/umd'],
+        frame: ['./src/frame/js/utils/polyfills', './src/frame/js/index']
     };
 
     if (options.hotComponents && !options.assetsOnly) {
-        entry.smooch.unshift('webpack-hot-middleware/client');
+        // entry.host.unshift('webpack-hot-middleware/client');
```
mspensieri: uncomment or remove?

make-webpack-config.js
```diff
@@ -19,41 +19,80 @@ module.exports = function(options) {
...
+                path.resolve(__dirname, 'test')
+            ]
+        },
+        'png|jpg|jpeg|gif|svg': {
+            loader: 'url-loader',
+            options: {
+                limit: fileLimit
+            }
         },
-        'json': 'json-loader',
```
mspensieri: json removed on purpose? Is it default in webpack 2?

src/frame/js/components/widget.jsx
```diff
@@ -25,7 +24,7 @@ export class WidgetComponent extends Component {
         smoochId: PropTypes.string,
         app: PropTypes.object.isRequired,
         settings: PropTypes.object.isRequired,
-        ui: PropTypes.object.isRequired,
```
mspensieri: This was unused I guess?

src/host/js/smooch.js
```diff
@@ -0,0 +1,93 @@
...
+            hostStyles.use();
+        });
+        return {
+            then: (next) => pendingInitNext = next,
+            catch: (next) => pendingInitCatch = next
+        };
+    }
+};
+
+window.__onLibReady = function onSmoochReady(Lib) {
```
mspensieri: Should this have a more specific name since it's in the global namespace? Something related to smooch?

src/host/js/utils/enquire.js
```diff
@@ -0,0 +1,29 @@
+import enquire from 'enquire.js';
```
mspensieri: What's the size of this lib? Is it lazy loaded at init time, or will this affect page load speed?

src/stylesheets/reset.less
mspensieri: Why do we need this at all if we're in an iframe? Isn't style already reset? Or is this meant to account for browser inconsistencies?

make-webpack-config.js
lemieux: forgot that one

make-webpack-config.js
lemieux: became default in webpack 2

src/frame/js/components/widget.jsx
lemieux: yeah

src/host/js/smooch.js
lemieux: yeah I'm even thinking about generating it randomly from the host and pass it to the iframe,

src/host/js/utils/enquire.js
lemieux: the lib is 2.5kb. This whole thing will be combined with a new way of loading the load with an async script loader like what GA is doing so page load is not affected too much by our lib.

src/stylesheets/reset.less
lemieux: that's exactly it, normalize.less from bootstrap is just about browser inconsistencies.

src/host/js/smooch.js
lemieux: I'll fix that in a future PR where I'll tackle packaging and build

src/frame/js/utils/dom.js
```diff
@@ -1,9 +1,12 @@
 import isMobile from 'ismobilejs';
 
+// Rewrite the loaders to avoid the style-loader used in the host part
+import hostStyles from '!css-loader?modules!less-loader!../../../host/stylesheets/iframe.less';
```
lemieux: so great thing about using css modules is that webpack will generate a classname that won't clash with anything (`_2TELtk5nDKlQudVSivRjpt`). It's really perfect for styles in the host page.

You can then import the stylesheet and find the classname like this `hostStyles.locals.widgetOpened`

src/frame/js/smooch.jsx
```diff
@@ -0,0 +1,390 @@
+import React from 'react';
+import { render as reactRender, unmountComponentAtNode } from 'react-dom';
+import pick from 'lodash.pick';
+import { batchActions } from 'redux-batched-actions';
+import { Provider } from 'react-redux';
+
+import '../stylesheets/main.less';
```
dannytranlx: Wait what? How does this work?

lemieux: eheh, in the webpack config, less files are configured to use the style-loader which automatically inject the stylesheet in the page when required.

PR #540, opened 2017-06-05
New build process

lemieux:
New version, new build process.

Here's the summary : 
- Cloudfront invalidation isn't required anymore because all files uploaded to S3 will have a different name
- The npm lib will only have the host part in it and will load the frame part from the CDN.
- Two sets of babel configs to try to minimize bundle sizes
- hot-loader is gone, it was unusable and was making the browser crash inside the iframe
- npm lib has no dependencies anymore. The host lib only has enquire.js as a dependency but the new entry point bundles it.

I tested the npm version on a SPA I had and also tried the cdn version on a dummy cdn.


make-webpack-config.js
```diff
@@ -88,31 +121,45 @@ module.exports = function(options) {
...
+        library: (options.npmRelease || options.frameOnly) ? undefined : 'Smooch',
+        libraryTarget: options.npmRelease ? 'commonjs2' : 'var',
         pathinfo: options.debug
     };
 
     const excludeFromStats = [
         /node_modules[\\\/]/
     ];
 
+    const baseFrameFilename = (options.frameOnly || options.hostOnly) ? `frame.${VERSION}` : 'frame';
```
alavers: /scratches head

package.json
```diff
@@ -1,156 +1,144 @@
...
-  "peerDependencies": {
-    "babel-polyfill": ">=6.9.0 <7.0",
-    "babel-runtime": ">=6.9.0 <7.0",
-    "react": "15.x.x",
-    "react-dom": "15.x.x"
-  },
-  "browser": {
-    "faye": "faye/browser/faye-browser"
-  }
+    "name": "smooch",
```
alavers: I believe NPM auto-formats your package.json to 2 space indent. Unless that's changed, can we restore it to 2 space indent to prevent it from always going back and forth?

make-webpack-config.js
lemieux: I should definitely add comments in that config file.

package.json
lemieux: yeah that was a mistake to auto-format this file. I'll fix it.

lemieux:
I need to write a guide about the release process... even for me it's a bit confusing when I didn't touch it in a while.

PR #541, opened 2017-06-06
Vendor in react-stripe-checkout

lemieux:
(will change base branch when the other is merged)

So this PR brings back Stripe functionnality in the mix. This will be pretty much the only part being injected in the host page from the frame lib. The reason why is simple : the modal need to appear in the host page and not the iframe.

To make it happen, I vendored in react-stripe-checkout that we were using to allow us to pass an execution context to it (the host page context). I also had to fix what looks like a bug in how StripeCheckout.configure works where `{} instanceof Object` would return false for some reason.

This is pretty much what I did back in the PoC.


hebime:
So this allows the modal to pop over the entire page instead of the just the little messenger window?

lemieux:
Yeah exactly. If we don't pass a context to react-stripe-checkout, it would just inject itself in the iframe.

src/frame/js/components/action.jsx
```diff
@@ -140,7 +140,8 @@ export class ActionComponent extends Component {
                                        currency={ currency.toUpperCase() }
                                        name={ stripeAccount.appName }
                                        image={ stripeAccount.iconUrl }
-                                       closed={ this.onStripeClose }>
+                                       closed={ this.onStripeClose }
+                                       executionContext={ parent }>
```
hebime: Parent here I'm guessing is the actual page itself? This is defined in react somewhere?

lemieux: `parent` is something available when inside an iframe. It refers to the context of the parent window.

mspensieri:
> To make it happen, I vendored in react-stripe-checkout that we were using to allow us to pass an execution context to it (the host page context). I also had to fix what looks like a bug in how StripeCheckout.configure works where {} instanceof Object would return false for some reason.

Based on this I guess you made modifications to the lib? If so, then maybe it should be two separate commits: 
1. Vendored stripe checkout lib
2. Modified vendored stripe checkout lib

so that we can clearly see where the edits are in the git history. From the diff of this PR it's really not clear which parts were modified

lemieux:
@mspensieri right, I'll keep this in mind for next time. Here are the lines that changed : 

https://github.com/smooch/smooch-js/pull/541/files#diff-85181ece056e5b0b15cfa09933733a17R19
https://github.com/smooch/smooch-js/pull/541/files#diff-85181ece056e5b0b15cfa09933733a17R333
https://github.com/smooch/smooch-js/pull/541/files#diff-85181ece056e5b0b15cfa09933733a17R404
https://github.com/smooch/smooch-js/pull/541/files#diff-85181ece056e5b0b15cfa09933733a17R461


does it do the trick?

mspensieri:
Kind of, it doesn't really show the diff of what changed though (were those lines added or modified from the original?). I would suggest rebase this pr into two commits, but your call if you want to follow my suggestion or not lol it's not _that_ big of a deal

lemieux:
I'm trying to not go redo the commits :p 

But the gist of it is the addition of the `executionContext` prop and then replacement of `document.something` with the `document`that is in `executionContext`.

The other change is in `configure` where the button id is passed.

The rest is identical to the latest version of the lib.

PR #542, opened 2017-06-07
Basic code splitting & optimization pass

lemieux:
Here's a first quick pass at optimizing the frame lib size. It was around 650kb minified non-gzip before the changes. The bundle was including the stylesheet too.

Here's what was optimized : 
- Stylesheet has been externalized and is loaded on its own.
- lodash.bindall has been removed since it was really big for no real reasons. Has been reimplemented as a util
- babel-env preset was configured to use built-ins to avoid duplication between the polyfill and what babel generates (https://github.com/babel/babel-preset-env#usebuiltins)
- webpack config for tree shaking

Current lib weights 685.5KB.
New one : 
Host : 16KB
Frame JS : 555KB
Frame CSS : 61.2KB

(all sizes are minified and non-gzipped)

One advantage of splitting the JS and CSS in the frame is that they can be fetched in parallel by the browser.


lemieux:
I tried to use webpack's code splitting to avoid loading the full widget if it's still closed and it worked... until I tried it in IE10... there was a problem where the babel-polyfill wasn't loaded first. I'll probably look into it later and include the polyfill directly in the iframe html. It was bringing the frame.js file down to ~430KB.

make-webpack-config.js
```diff
@@ -170,27 +175,37 @@ module.exports = function(options) {
...
     const plugins = [
         new webpack.DefinePlugin({
-            FRAME_LIB_URL: `'${publicPath}${frameLibFilename}'`
+            FRAME_JS_URL: `'${publicPath}${frameJsFilename}'`,
+            FRAME_CSS_URL: `'${publicPath}${frameCssFilename}'`
         })
     ];
 
+    if (buildType === 'frame') {
+        plugins.push(new ExtractTextPlugin(`frame.${VERSION}.css`));
```
mspensieri: is `.min.css` not a thing?

src/frame/js/utils/functions.js
```diff
@@ -0,0 +1,3 @@
+export function bindAll(context, ...methodNames) {
```
mspensieri: lol so big

make-webpack-config.js
lemieux: yes and no.. we do have a "minification" pass where we strip blank lines and comments... but I still call it .css since it still is readable... /shrug

src/frame/js/utils/functions.js
lemieux: lodash.bindall was like 5kb on the minified bundle.. made no sense to keep that

make-webpack-config.js
mspensieri: Is it possible to uglify? Do we get any space savings from that?

lemieux: you can't really uglify css since you have to have the css rules written in something the browser understands and your class names need to match.

the optimization pass is supposed to merge rules together if there have the same selector and to stripe any unecessary chars. you get your space savings from that. after that, there's gzip for more size reduction when it's transferred.

make-webpack-config.js
```diff
@@ -130,11 +134,12 @@ module.exports = function(options) {
             path.resolve(__dirname, 'src/frame/'),
             path.resolve(__dirname, 'src/shared/')
         ],
-        use: [
-            'style-loader',
-            'css-loader',
-            'less-loader'
-        ]
+        use: ExtractTextPlugin.extract({
```
wmora: why is this needed?

lemieux: this is how you tell webpack to generate and external file for this type of module. In this case, we're telling it to generate a css file for css modules it encounters instead of inlining it inside the js bundle. You could do the same for html files if you wanted for instance.

PR #543, opened 2017-06-08
Add Sentry support

lemieux:
This adds Sentry support for the frame part of the widget. The frame runtime is isolated from the host page which means that we wouldn't get any errors from the host. Also, now that the npm version will download the frame lib from the cdn, all of our new installs will have error capture support and will help us track and improve the widget proactively.

The deployment will now require a SENTRY_DSN and a SENTRY_AUTH_TOKEN (already set in Circle so we don't forget at that time) as described in https://docs.sentry.io/learn/cli/configuration/.

The changes to the deployment process are pretty simple, it creates a new release in Sentry, associates the commits with the release, uploads the sourcemap, and finalizes the release (as documented here https://docs.sentry.io/learn/cli/releases/).

I also removed the staging deployment config as it will require a different process anyway with the upcoming changes.

### DIFF
--- a/src/frame/js/services/core.js
+++ b/src/frame/js/services/core.js
@@ -1,8 +1,6 @@
 import { Smooch } from 'smooch-core/lib/smooch';
 import urljoin from 'urljoin';
 
-import { VERSION } from '../../../shared/js/constants/version';
-
 export function core({auth, appState}) {
     return new Smooch(auth, {
         serviceUrl: urljoin(appState.serverURL, 'v1'),
--- a/src/frame/js/smooch.jsx
+++ b/src/frame/js/smooch.jsx
@@ -3,6 +3,7 @@ import { render as reactRender } from 'react-dom';
 import pick from 'lodash.pick';
 import { batchActions } from 'redux-batched-actions';
 import { Provider } from 'react-redux';
+import Raven from 'raven-js';
 
 import '../stylesheets/main.less';
 
@@ -215,6 +216,14 @@ export function login(userId = '', jwt, attributes) {
             }
         }
     })).then((loginResponse) => {
+        Raven.setUserContext({
+            id: loginResponse.appUser.userId || loginResponse.appUser._id
+        });
+
+        Raven.setExtraContext({
+            appToken
+        });
+
         const actions = [];
         actions.push(userActions.setUser(loginResponse.appUser));
         actions.push(setApp(loginResponse.app));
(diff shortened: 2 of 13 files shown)


src/frame/js/services/core.js
```diff
@@ -1,8 +1,6 @@
 import { Smooch } from 'smooch-core/lib/smooch';
 import urljoin from 'urljoin';
 
-import { VERSION } from '../../../shared/js/constants/version';
```
lemieux: VERSION is now injected directly by webpack... not sure why it wasn't that way in the first place (I think it was because npm lib was just babelified and not processed by webpack at all, now it is).

src/frame/js/smooch.jsx
```diff
@@ -216,6 +216,14 @@ export function login(userId = '', jwt, attributes) {
             }
         }
     })).then((loginResponse) => {
+        Raven.setUserContext({
+            id: loginResponse.appUser.userId || loginResponse.appUser._id
+        });
+
+        Raven.setExtraContext({
+            appToken
```
lemieux: we'll replace this with appId when it will be used. I thought I could do it right now, but we don't return the appId as part of the init response.

This will be useful to track who's getting errors and get in touch if necessary.

src/frame/js/smooch.jsx
```diff
@@ -216,6 +216,14 @@ export function login(userId = '', jwt, attributes) {
             }
         }
     })).then((loginResponse) => {
+        Raven.setUserContext({
+            id: loginResponse.appUser.userId || loginResponse.appUser._id
```
lemieux: I was hesitant to use `userId` in case it was an email.... should we just use `_id`?

mspensieri: I think so

src/frame/js/utils/raven.js
```diff
@@ -0,0 +1,11 @@
+import Raven from 'raven-js';
+
+if (SENTRY_DSN) {
+    Raven.config(SENTRY_DSN, {
+        release: VERSION
+    }).install();
+
+    window.onunhandledrejection = function(e) {
```
mspensieri: dat casing 👌

lemieux: ahah

src/frame/js/smooch.jsx
dannytranlx: Nice

PR #544, opened 2017-06-13
Bring back the tests

lemieux:
This is a big one. I'm sorry, but it couldn't really be split in smaller chunks without making any sense.

In the wake of my "upgrade all the deps" frenzy + moving files around, tests needed to be updated.

Most of the changes are linked to the usage of Babel-rewire to properly inject/replace dependencies within modules. Previously, we were stubbing 1, 2, or even 3 level further in order to make our tests work and it was a bit crappy. Now with Rewire, we can replace exactly what we need inside the context of a module (even a reference within the module itself : smooch.jsx where `init` calls `login` internally... now we can stub `login`).

Couple of things to note when you read tests : 
- pure non-thunk actions are just spied on and not stubbed
- thunk actions that aren't async are stubbed as `sandbox.stub().returns(() => null)`
- thunk actions that are async are stubbed as `sandbox.stub().returns(() => Promise.resolve())`

Those are patterns that are recurrent across the test files.


lemieux:
One thing why I needed to change the way we mock/stub action is because I updated redux and redux-mock-store and it doesn't accept action without a type prop. We were mocking most actions with empty objects before.

package.json
```diff
@@ -112,21 +113,20 @@
...
-    "redux-thunk": "2.1.0",
+    "react-redux": "5.0.5",
+    "react-spinkit": "3.0.0",
+    "redux": "3.6.0",
+    "redux-batched-actions": "0.2.0",
+    "redux-mock-store": "1.2.3",
+    "redux-thunk": "2.2.0",
     "semver": "4.3.6",
     "sentry-cli-binary": "1.13.3",
     "sinon": "2.0.0-pre",
```
mspensieri: If you update this to 2.1 you can get rid of `sinon-as-promised` and use fancy new stuff like `callsFake`

package.json
```diff
@@ -112,21 +113,20 @@
...
-    "react-addons-test-utils": "15.1.0",
-    "react-document-title": "2.0.1",
-    "react-dom": "15.1.0",
+    "react": "15.5.4",
+    "react-addons-css-transition-group": "15.5.2",
+    "react-addons-test-utils": "15.5.1",
+    "react-document-title": "2.0.3",
+    "react-dom": "15.5.4",
     "react-hot-loader": "1.3.0",
     "react-imageloader": "2.1.0",
```
mspensieri: Seems to be vendored in now, remove?

Also, why the vendoring?

test/specs/services/faye.spec.js
```diff
@@ -32,26 +31,61 @@ const sandbox = sinon.sandbox.create();
...
-        sandbox.stub(conversationActions, 'incrementUnreadCount');
-        sandbox.stub(conversationActions, 'resetUnreadCount');
-        sandbox.stub(userActions, 'setUser');
-        sandbox.stub(fayeActions, 'setFayeConversationSubscription');
-        sandbox.stub(fayeActions, 'setFayeUserSubscription');
-
-        sandbox.stub(utilsDevice, 'getDeviceId').returns(123);
+
+        getMessagesStub = sandbox.stub().returns(() => Promise.resolve());
+        FayeRewire('getMessages', getMessagesStub);
```
mspensieri: So let me see if I understand what's happening here, because I was a bit confused at first glance. 

We're testing faye in this file
The faye module imports a bunch of symbols from other places
`FayeRewire` is a sort of dictionary of all the symbols imported by the Faye module
In this file, you import the same symbols that Faye is importing, but then hot-swap the reference _inside the faye module_ with a stubbed version?

So it's not working like a traditional mock where you replace the implementation of a function, what you're really replacing is the faye module's internal reference to that function. Is that right?

Basically I was confused how you could change stubs on different modules to instead all point to `FayeRewire`. I was expecting `ConversationActionsRewire` + `UserActionsRewire` + `ConversationServiceRewire`, etc.

test/specs/components/action.spec.js
```diff
@@ -83,19 +82,23 @@ describe('Action Component', () => {
             className: 'mockedLoading'
         });
 
-        sandbox.stub(userService, 'immediateUpdate');
-        userService.immediateUpdate.resolves();
 
-        sandbox.stub(conversationService, 'postPostback');
-        conversationService.postPostback.resolves();
+        immediateUpdateStub = sandbox.stub().returns(() => Promise.resolve());
```
mspensieri: Man, having `.returns(() => Promise.resolve())` everywhere is tedious. I wonder if we could write a sinon extension or something to add a new `returnsResolvingFn` or something like that?

package.json
lemieux: it was throwing warnings and errors with the upgrade of React and the repo is pretty much dead. https://github.com/hzdg/react-imageloader

test/specs/services/faye.spec.js
lemieux: > but then hot-swap the reference inside the faye module with a stubbed version?
Yeah that's it and from what I understand, you can't to rewire 2 levels deeper. It only work for that module directly (maybve I'm wrong, but it looked like that when I tested it).

> traditional mock where you replace the implementation of a function, what you're really replacing is the faye module's internal reference to that function. Is that right?

if you look at what jest does (https://facebook.github.io/jest/), it stubs by default any dependencies of a module and you have to tell it which ones to let go through unstubbed. imo it makes a much more isolated test case without potential side effects if you were to mock 3 levels down.

test/specs/components/action.spec.js
lemieux: yeah I'll check that out

package.json
lemieux: interrrresting, I'll update

test/specs/components/action.spec.js
```diff
@@ -83,19 +82,23 @@ describe('Action Component', () => {
...
+        ActionRewire('immediateUpdate', immediateUpdateStub);
```
dannytranlx: How does this rewire thing work again?

dannytranlx: Oh there's another thread, my b

lemieux:
@mspensieri here you go, I created sinon plugins for `returnsSyncThunk` and `returnsAsyncThunk`. I think I replaced them everywhere.

PR #545, opened 2017-06-14
Script loader

lemieux:
This PR adds the script loader that will be used by integrators.

I added a `npm run uglify:loader` that will output the script to be pasted in the docs. This is the final output : 

```
(function(e,n,t,o){var r;var s=[];var a;e[t]={init:function(){r=arguments;return{then:function(e){a=e}}},on:function(){s.push(arguments)}};e.__onSmoochHostReady__=function n(o){delete e.__onSmoochHostReady__;e[t]=o;if(r){var i=o.init.apply(o,r);if(a){i.then(a)}}for(var c=0;c<s.length;c++){o.on.apply(o,s[c])}};function i(){try{if(this.response.url){var e=n.getElementsByTagName("script")[0];var t=n.createElement("script");t.async=true;t.src=this.response.url;e.parentNode.insertBefore(t,e)}}catch(e){}}var c=new XMLHttpRequest;c.addEventListener("load",i);c.open("GET","https://"+o+".webloader.smooch.io/",true);c.responseType="json";c.send()})(window,document,"Smooch","<app-id>");
```

People will have to replace `<app-id>` with their appId and if for some reason they need the global var to be anything else than `Smooch` they can change it too (useful for future white labeling).

The way I set up the dev server, it reads what is in `src/loader/index.js` and put it in the page. By default, it's minified, but you can pass `minifyLoader: false` in your config to have it unminified (for debugging purposes).

### DIFF
--- a/dev-server/server.js
+++ b/dev-server/server.js
@@ -3,7 +3,7 @@ module.exports = function(options) {
     const express = require('express');
     const path = require('path');
     const fs = require('fs');
-    const uglifyJs =  require('uglify-js');
+    const uglifyJs = require('uglify-js');
 
     // require the page rendering logic
     const Renderer = require('./SimpleRenderer.js');
@@ -14,7 +14,7 @@ module.exports = function(options) {
         .replace(
             '\'https://\' + appId + \'.webloader.smooch.io/\'',
             '\'/webloader\''
-        );
+    );
 
     try {
         Object.assign(config, require('../config/config.json'));
@@ -64,14 +64,16 @@ module.exports = function(options) {
 
     app.get('/webloader', function(req, res) {
         res.json({
-            url:  '/_assets/smooch.js'
+            url: '/_assets/smooch.js'
         });
     });
 
     // application
     app.get('/*', function(req, res) {
-        if(config.minifyLoader) {
-            loaderScriptContent = uglifyJs.minify(loaderScriptContent, {compress: false}).code;
+        if (config.minifyLoader) {
+            loaderScriptContent = uglifyJs.minify(loaderScriptContent, {
+                compress: false
+            }).code;
         }
 
         const renderer = new Renderer({
--- a/dev-server/server.js
+++ b/dev-server/server.js
@@ -3,6 +3,8 @@ module.exports = function(options) {
     const express = require('express');
     const path = require('path');
     const fs = require('fs');
+    const uglifyJs =  require('uglify-js');
+
     // require the page rendering logic
     const Renderer = require('./SimpleRenderer.js');
 
@@ -68,6 +70,10 @@ module.exports = function(options) {
 
     // application
     app.get('/*', function(req, res) {
+        if(config.minifyLoader) {
+            loaderScriptContent = uglifyJs.minify(loaderScriptContent, {compress: false}).code;
+        }
+
         const renderer = new Renderer({
             loaderScript: loaderScriptContent,
             data: config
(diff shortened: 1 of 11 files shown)


make-webpack-config.js
```diff
@@ -79,6 +79,9 @@ module.exports = function(options) {
             path.resolve(__dirname, 'src/host/'),
             path.resolve(__dirname, 'src/shared/')
         ],
+        exclude: [
+            path.resolve(__dirname, 'src/host/js/umd.js')
```
lemieux: This is to solve a problem where webpack was injecting both AMD and CommonJS utility in the generated script so it was basically rendering the umd loader useless

dev-server/server.js
```diff
@@ -54,10 +62,20 @@ module.exports = function(options) {
...
 
+    app.get('/webloader', function(req, res) {
+        res.json({
+            url:  '/_assets/smooch.js'
+        });
+    });
+
     // application
     app.get('/*', function(req, res) {
+        if(config.minifyLoader) {
```
mspensieri: jsfmt

src/loader/index.js
```diff
@@ -0,0 +1,57 @@
...
+                    next = _next;
+                }
+            };
+        },
+        on: function() {
+            onCallArgs.push(arguments);
+        }
+    };
+
+    w.__onSmoochHostReady__ = function onHostReady(Lib) {
```
mspensieri: Should this be called `__on${globalVarName}HostReady__` ?

src/loader/index.js
```diff
@@ -0,0 +1,57 @@
...
+                firstTag.parentNode.insertBefore(tag, firstTag);
+            }
+        }
+        catch (e) {}
+    }
+
+    // let's make a request to know which version of the lib to load
+    var req = new XMLHttpRequest();
+    req.addEventListener('load', onLoad);
+    req.open('GET', 'https://' + appId + '.webloader.smooch.io/', true);
```
mspensieri: no way to configure this url in the params?

src/loader/index.js
lemieux: This variable is hardcoded in the host lib so it can't really be configurable. I tried to find a way to make it dynamic but I didn't find any. The host lib needs something to notify when it's loaded.

I don't think it's too bad, under the hood a lot of stuff is still branded Smooch, but the public facing variable name is configurable.

src/loader/index.js
lemieux: yeah we could have it as a param, but I felt it wasn't something that would really change much (if you don't count the dev server hack I had to do).

mspensieri: It won't need to change in the white labelling case?

lemieux: Yeah but I figured we'd provide a custom loader for white label customers... now... why can you customize the variable name then..? Well I figure it could be a use case for normal users to just rename that variable, but the URL is pretty much set in stone.

src/host/js/umd.js
```diff
@@ -4,12 +4,18 @@ import Smooch from './smooch';
...
-            return (root.Smooch = factory());
+            return factory();
         });
     } else if (typeof module === 'object' && module.exports) {
         module.exports = factory();
     } else {
-        root.Smooch = factory();
+        // in this case, the host lib must be used in conjunction with
+        // the script loader
+        if(root.__onSmoochHostReady__) {
```
dannytranlx: whitspace formatting :3

PR #546, opened 2017-06-15
Deployment updates

lemieux:
This adds the new deployment steps with for script loader. I replace grunt stuff with direct calls to the aws cli (it's available on Circle). I also split the umd loader in 3 different builds for convenience (and also, uglifyJS stopped liking our umd.js file for some reason.

tl;dr; when deploying there's an extra step that uploads a json file with the new script url in it in the webloader S3 bucket and then issues an invalidation request to cloudfront.

I tested it all with a whole dummy infra to replicate prod.


make-webpack-config.js
```diff
@@ -150,7 +153,7 @@ module.exports = function(options) {
         options.publicPath :
         buildType === 'dev' ?
             '/_assets/' :
-            'https://cdn.smooch.io/';
+            'https://cdn.smooch.rocks/';
```
mspensieri: leftover from testing?

make-webpack-config.js
```diff
@@ -169,7 +172,7 @@ module.exports = function(options) {
         filename: baseFilename + fileExtension,
         chunkFilename: buildType === 'dev' ? '[id].js' : '[chunkhash].js',
         sourceMapFilename: '[file].map',
-        library: buildType === 'host' ? 'Smooch' : undefined,
+        library: undefined,
```
mspensieri: ?

make-webpack-config.js
lemieux: oh..

make-webpack-config.js
lemieux: fixed

lemieux: Webpack doesn't need to expose a global variable anymore, it's handled by the script loader.

PR #547, opened 2017-06-19
Add handling for faye channel link failure

jugarrit:
This PR adds support for the new faye event that will be fired in new versions of the app.

### DIFF
--- a/src/js/services/integrations.js
+++ b/src/js/services/integrations.js
@@ -2,13 +2,39 @@ import { batchActions } from 'redux-batched-actions';
 
 import { core } from './core';
 
-import { setError, unsetError, setWeChatQRCode, setWeChatError, unsetWeChatError, setTwilioIntegrationState, resetTwilioIntegrationState, setViberQRCode, setViberError, unsetViberError, setLinkCode, setTransferRequestCode } from '../actions/integrations-actions';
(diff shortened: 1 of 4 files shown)


src/js/services/integrations.js
```diff
@@ -2,14 +2,40 @@ import { batchActions } from 'redux-batched-actions';
...
-import { setError, unsetError, setWeChatQRCode, setWeChatError, unsetWeChatError, setTwilioIntegrationState, resetTwilioIntegrationState, setViberQRCode, setViberError, unsetViberError, setLinkCode, setTransferRequestCode } from '../actions/integrations-actions';
+import { setError, unsetError, setWeChatQRCode, setTwilioIntegrationState, resetTwilioIntegrationState, setViberQRCode, setTransferRequestCode } from '../actions/integrations-actions';
 import { handleConversationUpdated } from './conversation';
 import { getUserId } from './user';
 import { updateUser } from '../actions/user-actions';
 
 let fetchingWeChat = false;
 let fetchingViber = false;
 
+function handleLinkFailure(dispatch, getState, error) {
```
lemieux: hmm I think this should be transform in an action.

```
function handleLinkFailure(error) {
    return (dispatch, getState) => { ... };
}
```

and you do `dispatch(handleLinkFailure(error))` in other places.

It feels more redux-y this way.

jugarrit: Agreed, I forgot how to redux.

PR #557, opened 2017-07-04
Update some build dependencies and clean up grunt

lemieux:
I tried a clean install of the project and some dependencies were missing and/or out of date. I also cleaned up the Grunt config since some tasks were obsolete for us.


.gitignore
```diff
@@ -17,3 +17,4 @@ example/*
 !example/template/*.html
 npm-debug*
 package-lock.json
+.env
```
lemieux: this file is where I put env var... I use `source .env` before running commands.. you won't usually need this unless you are testing deployments.

PR #558, opened 2017-07-06
Fix sourcemap generation

lemieux:
This should fix the problem where sourcemaps were empty.


lemieux:
Yeah I think so, couple of things changed between webpack 1 and 2. That must have been one of them.

PR #560, opened 2017-07-06
Fix channel page sizing

lemieux:
The page was at the right size, but without top 0, it wasn't at the right place. I set top 0 and a padding a the top to clear the header and it's fine now.


lemieux:
image

lemieux:
image

PR #561, opened 2017-07-07
Fix iframe sizing in embedded mode

lemieux:
What the title says.


src/host/js/smooch.js
```diff
@@ -135,9 +135,11 @@ function onSmoochReady(_Lib) {
 
 function injectFrame() {
     if (!iframe) {
+        let loaded = false;
         iframe = document.createElement('iframe');
+
+        iframe.frameborder = 0;
```
lemieux: This is just to prevent the border to show up until the stylesheet is applied. I

PR #566, opened 2017-07-10
Fix jumpy input on ios

lemieux:
Couple of small fixes for iOS

### DIFF
--- a/src/host/stylesheets/iframe.less
+++ b/src/host/stylesheets/iframe.less
@@ -77,13 +77,12 @@
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/host/stylesheets/iframe.less
```diff
@@ -77,13 +77,12 @@
         }
     }
 
-    .widgetOpened {
-        body,
-        html {
+    html.widgetOpened {
```
lemieux: The actual fix for the jumpy input is here... scope wasn't applied correctly on the page.

PR #572, opened 2017-07-21
First pass at Web SDK refactoring

lemieux:
So first, sorry if this PR looks like a mess, but I had to start somewhere.

At this point, I tried to refactor the init flow for anon users. 

The widget : 
-  doesn't work for authenticated users
-  can't send messages yet.
- etc.

tl;dr; it doesn't work.

So what changed? 

I refactored the reducers a bit to make it cleaner.. no more `app` reducer, a new `config` one was created and will hold all the config returned by the server + 1-2 client side config flags (imageUploadEnabled for instance). The server response for config also changed, so stuff like `app.settings.web` is now located under `config.style` for instance. Another big changes, since the new endpoints accept a `sessionToken` auth, we can't rely on `smooch-core` anymore.. so I've been porting http calls over and the `http` utils is an action now. It's a bit smarter than the `http` util in `core` since it has access to auth values in the store and all that stuff so you don't have to specify it when calling it.

I made sure to fix all the tests to work with the new structure and I commented parts of the code I wasn't sure they would stay with the new flow or not.

Again, really sorry, this is a mess, but new endpoints aren't there yet so I'm relying on tests to work.


src/frame/js/actions/conversation.js
```diff
@@ -240,10 +239,13 @@ export function sendMessage(props) {
 
 export function postPostback(actionId) {
     return (dispatch, getState) => {
-        return core(getState()).conversations.postPostback(getUserId(getState()), actionId)
-            .catch(() => {
-                dispatch(showErrorNotification(getState().ui.text.actionPostbackError));
-            });
+        const {user: {_id}, config: {appId}} = getState();
+
+        return dispatch(http('POST', `/apps/${appId}/appusers/${_id}/conversation/postback`, {
```
mspensieri: The path is just `/:id/postback`, no need for `/conversation` anymore

src/frame/js/actions/conversation.js
```diff
@@ -88,33 +90,36 @@ const throttlePerUser = (userId) => {
...
-        return core(getState()).appUsers.uploadImage(getUserId(getState()), blob, {
-            role: 'appUser',
-            deviceId: getDeviceId()
+        const data = new FormData();
+        data.append('role', 'appUser');
+        data.append('deviceId', getDeviceId());
+        data.append('source', blob);
+
+        Object.keys(message).forEach((key) => {
+            data.append(key, message[key]);
```
mspensieri: 🤔  What other fields are they allowed to send here?

src/frame/js/actions/conversation.js
```diff
@@ -88,33 +90,36 @@ const throttlePerUser = (userId) => {
 
 function postSendMessage(message) {
     return (dispatch, getState) => {
-        return core(getState()).appUsers.sendMessage(getUserId(getState()), message);
+        const {user: {_id}} = getState();
+        return dispatch(http('POST', `/appusers/${_id}/messages`, {
+            ...message,
+            role: 'appUser',
+            deviceId: getDeviceId()
```
mspensieri: Obviously this wasn't decided when the PR was written, but this will change into `sender.client.id`. Can be done in a separate PR of course

src/frame/js/actions/conversation.js
```diff
@@ -516,3 +514,56 @@ export function disconnectFaye() {
...
+export function startConversation() {
+    return (dispatch, getState) => {
+        const {user: {_id: userId, pendingAttributes}, config: {appId}, conversation: {_id: conversationId}} = getState();
+
+        if (conversationId) {
+            return Promise.resolve();
+        }
+
+        return dispatch(http('POST', userId ? `/appusers/${userId}/conversations` : `/apps/${appId}/appusers`, {
+            ...pendingAttributes,
```
mspensieri: Should these be under an `appUser` field? Especially if the path is `/appusers/${userId}/conversations`, putting values like `firstName` and `properties` at the root seems weird

src/frame/js/smooch.jsx
```diff
@@ -209,13 +220,9 @@ export function login(userId = '', jwt, attributes) {
             id: loginResponse.appUser.userId || loginResponse.appUser._id
         });
 
-        Raven.setExtraContext({
-            appToken
```
mspensieri: don't want to change this to appId?

src/frame/js/actions/conversation.js
lemieux: :+1:

src/frame/js/actions/conversation.js
lemieux: hmm good question, I just ported whatever we had in smooch-core

src/frame/js/actions/conversation.js
lemieux: :+1:

src/frame/js/actions/config.js
```diff
@@ -0,0 +1,43 @@
...
+    return {
+        type: RESET_CONFIG
+    };
+}
+
+export function fetchConfig() {
+    return (dispatch, getState) => {
+        const {configBaseUrl, appId} = getState().config;
+
+        return dispatch(http('GET', `/client/apps/${appId}/config`, {}, {}, configBaseUrl))
```
wmora: `/client` is already provided in the config response, but I guess you'll change the endpoints in another PR

src/frame/js/actions/conversation.js
lemieux: yeah seems weird to me too, I followed the spec we had though, I recheck again.

src/frame/js/smooch.jsx
lemieux: yeah it's done here https://github.com/smooch/smooch-js/pull/572/files#diff-8f16faf8c766689cc06ac2f62eb9db67R129

mspensieri: Missed that, 👍

src/frame/js/actions/config.js
lemieux: they are provided by the config response (which is the call made here)... it's not provided for the config call though.

src/frame/js/actions/faye.js
```diff
@@ -155,9 +184,10 @@ export function updateUser(currentAppUser, nextAppUser) {
             disconnectFaye();
 
             return dispatch(subscribeUser()).then(() => {
-                if (nextAppUser.conversationStarted) {
-                    return dispatch(handleConversationUpdated());
-                }
+                // TODO : figure out if still relevant
+                // if (nextAppUser.conversationStarted) {
```
wmora: what's this for?

src/frame/js/actions/http.js
```diff
@@ -0,0 +1,97 @@
+import urljoin from 'urljoin';
+
+/**
+ * Stringifies query parameters and append them to the url
+ * @param  {string} url  - an url
+ * @param  {object} data - an object containing the query parameters
+ * @return {string}      - the final url
+ */
+export function stringifyGETParams(url, data) {
```
wmora: should we unit test this?

lemieux: yeah we have those, I just forgot to port them over from core https://github.com/smooch/smooch-core-js/blob/master/test/specs/utils/http.spec.js

src/frame/js/actions/http.js
```diff
@@ -0,0 +1,97 @@
...
+    if (response.status === 202 || response.status === 204) {
+        return Promise.resolve();
+    }
+
+    const contentType = response.headers.get('Content-Type') || '';
+    const isJson = contentType.indexOf('application/json') > -1;
+
+    if (response.status >= 200 && response.status < 300) {
+        return isJson ? response.json() : Promise.resolve();
+    } else {
```
wmora: nit: `else if (isJson)`?

src/frame/js/actions/faye.js
lemieux: this was for the linking for where User A has no conversation yet and links a channel and it gets merged with User B that has a conversation started. The widget would need to "init" itself with the new user state... I'm not sure this flow will still be relevant with the new stuff (or at least will be different).

src/frame/js/components/Action.jsx
```diff
@@ -195,12 +195,12 @@ export class ActionComponent extends Component {
...
 
-export default connect(({app, ui: {text}, user}) => {
+export default connect(({ui: {text}, user, config}) => {
     return {
         user,
         actionPaymentCompletedText: text.actionPaymentCompleted,
-        integrations: app.integrations,
-        stripe: app.stripe
+        integrations: config.integrations,
+        stripe: config.stripe
```
wmora: this is part of `config.integrations` now

src/frame/js/actions/http.js
lemieux: hmm yeah.. I didn't check the code when I copied it.

src/frame/js/components/Action.jsx
lemieux: it was still part of the config call when I implemented it... I'll fix this once the backend is more "official".. most of the component changes are just to be able to render the widget with the new changes

src/frame/js/components/ChatInput.jsx
```diff
@@ -132,10 +132,10 @@ export class ChatInputComponent extends Component {
     }
 }
 
-export default connect(({appState, app, ui, conversation: {unreadCount}}) => {
+export default connect(({ui, conversation: {unreadCount}, config}) => {
     return {
-        imageUploadEnabled: appState.imageUploadEnabled,
-        accentColor: app.settings.web.accentColor,
+        imageUploadEnabled: config.imageUploadEnabled,
```
wmora: (a bit lost here, sorry) Is this config the same as the response from `/config`? If so, where is `imageUploadEnabled` assigned to it?

lemieux: config is a mix of what the `/config` endpoint return and local config (it's set here https://github.com/smooch/smooch-js/pull/572/files#diff-8f16faf8c766689cc06ac2f62eb9db67R138)

src/frame/js/actions/http.js
lemieux: done

PR #573, opened 2017-07-26
Login flow refactor

lemieux:
This is a simple one that tackles the login flow, which doesn't do much aside setting the right props in the store.


src/frame/js/actions/auth.js
```diff
@@ -1,13 +1,30 @@
...
-export function login(props) {
+export function login() {
     return (dispatch, getState) => {
-        const {config: {appId}} = getState();
-        return dispatch(http('POST', `/apps/${appId}/login`, props));
+        const {config: {appId}, user: {userId}, auth: {sessionToken}} = getState();
+        return dispatch(http('POST', `/apps/${appId}/login`, {
+            userId,
+            sessionToken,
+            clientId: getDeviceId(appId)
```
mspensieri: Isn't login accepting the whole `client` object?

src/frame/js/actions/conversation.js
```diff
@@ -515,18 +517,31 @@ export function disconnectFaye() {
...
 
+        if (sessionToken) {
+            actions.push(setAuth({
+                sessionToken
+            }));
+        }
+
         dispatch(batchActions(actions));
+
+        if (appUser.conversationStarted) {
```
mspensieri: might be better to check `if(conversation)` since we have it?

src/frame/js/utils/device.js
```diff
@@ -2,8 +2,8 @@ import uuid from 'uuid';
 
 import storage from './storage';
 
-export function getDeviceId() {
-    const SK_STORAGE = 'sk_deviceid';
+export function getDeviceId(appId) {
+    const SK_STORAGE = `${appId}.clientId`;
```
mspensieri: How come client id is scoped per app now?

src/frame/js/actions/auth.js
lemieux: @hebime told me only clientId was fine.

src/frame/js/actions/conversation.js
lemieux: I guess we could, but that'd be a backend logic error if it returns `conversationStarted` for the user and not return the conversation with it.

src/frame/js/actions/auth.js
wmora: but isn't it `client.id`?

hebime: we take the entire client now 😞

src/frame/js/actions/conversation.js
```diff
@@ -90,23 +91,23 @@ const throttlePerUser = (userId) => {
 
 function postSendMessage(message) {
     return (dispatch, getState) => {
-        const {user: {_id}} = getState();
+        const {config: {appId}, user: {_id}} = getState();
         return dispatch(http('POST', `/appusers/${_id}/messages`, {
             ...message,
             role: 'appUser',
-            deviceId: getDeviceId()
+            deviceId: getDeviceId(appId)
```
wmora: (for future PR) this goes in a separate `sender.client` object now

wmora:
late 👍

PR #574, opened 2017-07-26
Fetch anonymous user + conversation flow

lemieux:
This covers the case where an appUserId and sessionToken is already persisted and need to fetch the user + convo.

It also contains couple of fixes for some unrelated stuff.

### DIFF
--- a/src/frame/js/utils/functions.js
+++ b/src/frame/js/utils/functions.js
@@ -1,3 +1,18 @@
 export function bindAll(context, ...methodNames) {
     methodNames.forEach((methodName) => context[methodName] = context[methodName].bind(context));
 }
+
+export function deepFreeze(obj) {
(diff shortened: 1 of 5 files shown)


src/frame/js/utils/functions.js
```diff
@@ -1,3 +1,18 @@
 export function bindAll(context, ...methodNames) {
     methodNames.forEach((methodName) => context[methodName] = context[methodName].bind(context));
 }
+
+export function deepFreeze(obj) {
```
mspensieri: unused??

lemieux: hmm yeah leftover of something I was trying to do... `Smooch.getConversation()` is returning the conversation object from the store and I wanted to deep freeze it before returning it to prevent tampering with the messages.. I'm thinking about freezing the store state after each actions though...

PR #575, opened 2017-07-27
Fetch user and conversation if user exists

lemieux:
This tackles the flow where the user and sessionToken were persisted locally and the widget fetches that information.


lemieux:
Oh well I already opened that PR... #574

PR #577, opened 2017-07-31
Port calls to API from Smooch Core

lemieux:
This PR removes the dependency to smooch-core and ports over all the api calls inside the SDK.


src/frame/js/actions/integrations.js
```diff
@@ -229,17 +261,14 @@ export function fetchMessageBirdAttributes() {
...
-            .then(({appUser}) => {
-                dispatch(updateUser(appUser));
-
-                if (appUser.conversationStarted) {
-                    return dispatch(handleConversationUpdated());
-                }
-            })
-            .then(() => {
+        const {config: {appId}, user: {_id, pendingClients}} = getState();
+        return dispatch(http('POST', `/client/apps/${appId}/appusers/${_id}/clients`, data))
```
mspensieri: `/client` 👀 

Shows up a bunch of times in the rest of this file too

src/frame/js/actions/stripe.js
```diff
@@ -1,18 +1,22 @@
...
-import { core } from '../utils/core';
-import { getUserId } from './user';
+import http from './http';
 
 export function createTransaction(actionId, token) {
     return (dispatch, getState) => {
-        return core(getState()).appUsers.stripe.createTransaction(getUserId(getState()), actionId, token).catch((e) => {
-            dispatch(showErrorNotification(getState().ui.text.actionPaymentError));
+        const {config: {appId}, user: {_id}, ui} = getState();
+        return dispatch(http('GET', `/client/apps/${appId}/appusers/${_id}/stripe/transaction`, {
```
mspensieri: should be a `POST`

src/frame/js/actions/user.js
```diff
@@ -46,13 +43,14 @@ export function immediateUpdate(props) {
...
-            return core(getState()).appUsers.update(getUserId(getState()), props).then((response) => {
-                dispatch(setUser(response.appUser));
-                if (updateToResolve) {
-                    updateToResolve(response);
-                }
-                return response;
-            });
+            return dispatch(http('PUT', `/apps/${appId}/appusers/${user._id}`, props))
+                .then((response) => {
+                    dispatch(setUser(response.appUser));
```
mspensieri: PUT user returns an empty body now

src/frame/js/actions/integrations.js
```diff
@@ -250,11 +279,17 @@ export function linkSMSChannel(userId, data) {
...
     return (dispatch, getState) => {
-        return core(getState()).appUsers.unlinkChannel(userId, type)
+        const {config: {appId}, user: {_id, clients, pendingClients}} = getState();
+        const client = clients.find((client) => client.platform === type);
+
+        if (!client) {
+            return Promise.resolve();
+        }
+
+        return dispatch(http('DELETE', `/client/apps/${appId}/appusers/${_id}/clients/${client._id}`))
```
mspensieri: clients are accessed by their `.id` not `._id`. In fact I don't think we return `._id` in the API at all for a client. We might want to discuss this though

src/frame/js/actions/integrations.js
lemieux: :man_facepalming:

src/frame/js/actions/integrations.js
```diff
@@ -327,40 +364,24 @@ export function failSMSLink(error, type) {
     };
 }
 
-export function fetchViberQRCode() {
+export function fetchTransferRequestCode(type) {
     return (dispatch, getState) => {
-        const {integrations: {viber}} = getState();
+        const {user: {_id}, config: {integrations, appId}} = getState();
+        const integration = integrations.find(({type:_type}) => type === _type);
```
alavers: Deconstruction run amok! Isn't `.find((integration) => type === integration.type);` a fair bit clearer?

lemieux: yeah... I was thinking that too when I read it back lol

PR #582, opened 2017-08-03
Stripe Refactor

lemieux:
This makes the widget use the appName and iconUrl from the config.

Also, I remove esformatter-semicolons because it wasn't working with jsx.

### DIFF
--- a/src/frame/js/components/Action.jsx
+++ b/src/frame/js/components/Action.jsx
@@ -15,6 +15,10 @@ import Loading from './Loading';
 export class ActionComponent extends Component {
 
     static propTypes = {
+        dispatch: PropTypes.func.isRequired,
+        _id: PropTypes.string.isRequired,
+        appName: PropTypes.string.isRequired,
+        appIconUrl: PropTypes.string.isRequired,
         text: PropTypes.string.isRequired,
         type: PropTypes.string,
         buttonColor: PropTypes.string,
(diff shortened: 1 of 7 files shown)


src/frame/js/components/Action.jsx
```diff
@@ -200,7 +201,9 @@ export default connect(({ui: {text}, user, config}) => {
         user,
         actionPaymentCompletedText: text.actionPaymentCompleted,
         integrations: config.integrations,
-        stripe: config.stripe
+        stripe: config.stripe,
```
mspensieri: still exists or no?

src/frame/js/components/Action.jsx
```diff
@@ -139,8 +140,8 @@ export class ActionComponent extends Component {
                                        email={ user.email }
                                        amount={ amount }
                                        currency={ currency.toUpperCase() }
-                                       name={ stripeAccount.appName }
-                                       image={ stripeAccount.iconUrl }
```
mspensieri: Is there a service or action that was fetching this data that should be deleted also?

src/frame/js/components/Action.jsx
lemieux: No it doesn't, info is in the config.integrations array now.

src/frame/js/components/Action.jsx
lemieux: Yes, this https://github.com/smooch/smooch-js/pull/582/commits/89faf6ba83303f89c2135857409b48bb8e79bf6e

src/frame/js/components/Action.jsx
mspensieri: kill this prop then?

lemieux: oh I thought it was dead

PR #583, opened 2017-08-04
Handle linking events where current user is the survivor

lemieux:
This brings back handling of the linking event in the case the linking didn't replace the current user.

The other case will be handled is in another PR.

I also add to fix some components to use the new store structure in order to make some part of the widget appear.

### DIFF
--- a/src/frame/js/smooch.jsx
+++ b/src/frame/js/smooch.jsx
@@ -175,20 +175,29 @@ export function init(props = {}) {
 
     return store.dispatch(fetchConfig())
         .then(() => {
-            if (props.userId && props.jwt) {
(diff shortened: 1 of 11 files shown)


src/frame/js/smooch.jsx
```diff
@@ -175,20 +175,29 @@ export function init(props = {}) {
...
+            const {config: {app: {status}}} = store.getState();
 
-            if (appUserId && sessionToken) {
-                return store.dispatch(fetchUserConversation());
-            }
-        })
-        .then(() => {
-            if (!props.embedded) {
-                render();
+            if (status !== 'active') {
```
mspensieri: 👍

PR #585, opened 2017-08-08
Handle link event when the appUser _id changes

mspensieri:
When a link completes, it's possible for an appUser's `_id` to change.
The SDK should be aware of this and be able to recover from such a
change. I've added two new pieces of logic:

1. When a `link` event comes through faye and the user _id changes,
   we change internal state to point to the new user, and
   re-initialize (via login if the user has a `userId`, or via a user
   fetch if the user is using a `sessionToken`)
2. When initializing the SDK with an anonymous user, if the server
   returns an `invalid_auth` error (e.g. if the stored appUserId or
   sessionToken are invalid), we clear any local state and allow
   execution to continue with a new anonymous user. This covers the
   case where an anonymous user has been merged out of existence via
   linking.

### DIFF
--- a/src/frame/js/actions/conversation.js
+++ b/src/frame/js/actions/conversation.js
@@ -516,7 +516,7 @@ export function handleUserConversationResponse({appUser, conversation, hasPrevio
             setUser(appUser),
             setConversation({
                 ...conversation,
-                hasMoreMessages: hasPrevious
+                hasMoreMessages: !!hasPrevious
             }),
             setMessages(messages)
         ];
--- a/src/frame/js/actions/faye.js
+++ b/src/frame/js/actions/faye.js
@@ -3,11 +3,12 @@ import { batchActions } from 'redux-batched-actions';
 
 import { setUser } from './user';
 import { showSettings, showTypingIndicator, hideTypingIndicator, hideChannelPage, hideConnectNotification } from './app-state';
-import { addMessage, incrementUnreadCount, resetUnreadCount, getMessages, disconnectFaye } from './conversation';
+import { addMessage, incrementUnreadCount, resetUnreadCount, getMessages, fetchUserConversation } from './conversation';
 import { cancelSMSLink, failSMSLink } from './integrations';
 import { getClientId } from '../utils/client';
 import { ANIMATION_TIMINGS } from '../constants/styles';
-
+import { login, setAuth } from './auth';
+import { setItem } from '../utils/storage';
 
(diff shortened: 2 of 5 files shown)


test/specs/smooch.spec.js
```diff
@@ -97,9 +105,11 @@ describe('Smooch', () => {
         });
 
         getItemStub = sandbox.stub();
+        removeItemStub = sandbox.stub();
         SmoochRewire('storage', {
```
alavers: Does `sandbox.restore` also reset these `__Rewire__`s somehow?

mspensieri: No clue

lemieux: No it doesn't, but iirc I put a global reset in a afterEach in the setup file.

mspensieri: https://github.com/smooch/smooch-js/blob/f/compat-expansion/test/index.js#L12

Go be on vacation

lemieux: Yes captain

src/frame/js/actions/conversation.js
```diff
@@ -516,7 +516,7 @@ export function handleUserConversationResponse({appUser, conversation, hasPrevio
             setUser(appUser),
             setConversation({
                 ...conversation,
-                hasMoreMessages: hasPrevious
+                hasMoreMessages: !!hasPrevious
```
hebime: Isn't `hasPrevious` already a bool?

mspensieri: In some responses we don't include `messages` or `hasPrevious`, which was leading to react throwing an error "this prop is required but undefined was passed". This makes sure to coerce `undefined` to `false`

src/frame/js/actions/faye.js
```diff
@@ -143,9 +144,30 @@ function handleActivityEvents(events) {
...
 
 function updateUser(currentAppUser, nextAppUser, client) {
-    return (dispatch) => {
+    return (dispatch, getState) => {
         if (currentAppUser._id !== nextAppUser._id) {
-            // wat do?
+            const {config: {appId}} = getState();
+            const actions = [];
+            actions.push(setUser({
+                ...currentAppUser,
```
lemieux: I think we should discard the current user props at this point.

mspensieri: At first I was keeping the `currentAppUser` props because some views were exploding from a lack of `clients` and `pendingClients` array, but I fixed that in a later PR, so in fact yes we might be able to remove this

PR #586, opened 2017-08-08
Add proper handling of switch user flows

mspensieri:
1. Fixed handling of `login` response, and clean up some state that
   was incorrectly preserved between users. Also fixed `login`
   incorrectly clearing the `jwt` after a successful call
2. Added full implementation of `logout`
3. Fix `startConversation` when a `userId` and `jwt` are supplied in
   `init`
4. Fix settings page when the user has not been created yet

### DIFF
--- a/src/frame/js/actions/auth.js
+++ b/src/frame/js/actions/auth.js
@@ -1,29 +1,78 @@
-import http from './http';
-import { handleUserConversationResponse } from './conversation';
+import { batchActions } from 'redux-batched-actions';
 
-import { getClientInfo } from '../utils/client';
+import http from './http';
+import { resetConversation, handleUserConversationResponse, disconnectFaye } from './conversation';
+import { setUser, resetUser } from './user';
+
+import { getClientId, getClientInfo } from '../utils/client';
 import { removeItem } from '../utils/storage';
 
 export const SET_AUTH = 'SET_AUTH';
 export const RESET_AUTH = 'RESET_AUTH';
 
-export function login() {
+export function login(userId, jwt) {
     return (dispatch, getState) => {
-        const {config: {appId}, user: {userId}, auth: {sessionToken}} = getState();
+        const {config: {appId}, auth: {sessionToken}} = getState();
+
+        const actions = [
+            setAuth({
+                jwt
+            }),
+            setUser({
+                userId
+            }),
+            resetConversation()
+        ];
+
+        dispatch(disconnectFaye());
+        dispatch(batchActions(actions));
+
         return dispatch(http('POST', `/apps/${appId}/login`, {
             userId,
             sessionToken,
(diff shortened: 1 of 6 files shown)


src/frame/js/actions/auth.js
```diff
@@ -1,33 +1,85 @@
...
+                    id: getClientId(appId)
+                }
+            }));
+        } else {
+            promise = Promise.resolve();
+        }
+
+        return promise.then(() => {
+            const actions = [
+                resetAuth(),
```
hebime: This is for `JWT` specifically? Do we treat sessionTokens differently by calling `removeItem(${appId}.sessionToken);`?

mspensieri: so `resetAuth()` removes the session token and JWT from memory (the redux store), `removeItem` clears the session token from the browser's local storage. We don't store JWTs in local storage, so we don't need to remove that

hebime: Bueno!

src/frame/js/actions/auth.js
```diff
@@ -1,33 +1,85 @@
...
+                jwt
+            }),
+            setUser({
+                userId
+            }),
+            resetConversation(),
+            resetIntegrations()
+        ];
+
+        dispatch(disconnectFaye());
```
alavers: Any reason why `disconnectFaye` isn't done as part of the batch?

src/frame/js/actions/auth.js
```diff
@@ -1,33 +1,85 @@
...
+export function logout() {
+    return (dispatch, getState) => {
+        const {config: {appId}, user: {_id: appUserId, userId}} = getState();
+
+        if (!userId) {
+            return Promise.resolve();
+        }
+
+        let promise;
+        if (appUserId) {
```
alavers: if you have a `userId` without a `appUserId` it's because you logged in but never sent a message... is that right?

src/frame/js/actions/auth.js
mspensieri: `batchActions` doesn't work for async actions like `disconnectFaye`. At least I couldn't get it to work

src/frame/js/actions/auth.js
mspensieri: It's because the SDK was initialized with a `userId`, but the user doesn't exist on the server yet. So yeah I guess that implies they haven't sent a message yet.

But it's also to have never sent a message and still have an `appUserId`, if the user was pre-created for example

src/frame/js/actions/auth.js
lemieux: Confirmed. Batched actions are not working with thunks.

alavers: Hey! Go back to 🍹

PR #587, opened 2017-08-08
Respect user profile config returned by the server

mspensieri:
1. Use `config.profile.uploadInterval` instead of hardcoded value for
   user props throttling
2. Do not upload user props if `config.profile.enabled` is false
3. Wait to upload user props if the user has not been created yet
4. Fix `isDirty` check being dirty too often
5. Upload any pending user props after a user is created
6. Fix `update` tests to mock `setTimeout` instead of relying on
   broken sinon timers

### DIFF
--- a/src/frame/js/actions/user.js
+++ b/src/frame/js/actions/user.js
@@ -33,7 +33,7 @@ export function immediateUpdate(props) {
         }
 
         lastUpdateAttempt = Date.now();
-        console.log('Updating attempt');
+
         props = pick(Object.assign({}, pendingUserProps, props), EDITABLE_PROPERTIES);
         pendingUserProps = {};
 
--- a/src/frame/js/actions/user.js
+++ b/src/frame/js/actions/user.js
@@ -1,13 +1,12 @@
 import deepEqual from 'deep-equal';
+import pick from 'lodash.pick';
 
 import http from './http';
-
 
 export const SET_USER = 'SET_USER';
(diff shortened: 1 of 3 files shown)


src/frame/js/actions/user.js
```diff
@@ -34,15 +33,14 @@ export function immediateUpdate(props) {
         }
 
         lastUpdateAttempt = Date.now();
-
-        props = Object.assign({}, pendingUserProps, props);
+        console.log('Updating attempt');
```
wmora: 🚫

mspensieri: 😱

src/frame/js/actions/user.js
```diff
@@ -96,6 +99,14 @@ export function update(props) {
     };
 }
 
+export function _resetState() {
```
lemieux: does it really need to be exported? if it needs to be called from the test, you can use the Rewire API to get it.

test/specs/actions/user.spec.js
```diff
@@ -47,34 +59,77 @@ describe('User Actions', () => {
...
+                };
+
+                return mockedStore.dispatch(userActions.immediateUpdate(props)).then(() => {
+                    httpStub.should.not.have.been.called;
+                });
+            });
         });
     });
 
     // skip these untils fake timers weirdness is resolved.
```
lemieux: comment should go away now

PR #588, opened 2017-08-09
Enable channel linking for users that have not started a convo

mspensieri:
Users that have not been created yet on the server should still be
allowed to link channels. This change forces user creation when a
linking action is initiated. Linking actions include generating a
transfer request, requesting a QR code for WeChat or Viber, or calling
the `POST /clients` to start SMS linking.

If a user has already been created, but the faye connection does not
exist for some reason, we instead force a connection to faye when a
linking action is initiated.

This change also modifies a few views that were refusing to render
themselves when the smooch user id did not exist, to make them
compatible with a non-existing user.

### DIFF
--- a/test/specs/actions/faye.spec.js
+++ b/test/specs/actions/faye.spec.js
@@ -342,11 +342,76 @@ describe('Faye Actions', () => {
     });
 
     describe('subscribe', () => {
-        it('should call setFayeSubcription', () => {
-            mockedStore = createMockedStore(sandbox, generateBaseStoreProps());
-            mockedStore.dispatch(fayeActions.subscribe()).then(() => {
-                setFayeSubscriptionSpy.should.have.been.calledOnce;
+        let getClientStub;
+        let subscribeStub;
+
+        beforeEach(() => {
+            subscribeStub = sandbox.stub().resolves();
+
+            getClientStub = sandbox.stub().returnsSyncThunk({
+                value: {
+                    subscribe: subscribeStub
+                }
             });
+            FayeRewire('getClient', getClientStub);
+        });
+
+        it('should abort if user does not exist', () => {
+            mockedStore = createMockedStore(sandbox, generateBaseStoreProps());
+
+            return mockedStore.dispatch(fayeActions.subscribe())
+                .then(() => {
+                    getClientStub.should.not.have.been.called;
+                    subscribeStub.should.not.have.been.called;
+                    setFayeSubscriptionSpy.should.not.have.been.called;
+                });
+        });
+
+        it('should return an existing subscription', () => {
+            const existingSubscription = Promise.resolve();
+
+            mockedStore = createMockedStore(sandbox, generateBaseStoreProps({
+                user: {
+                    _id: hat()
+                },
+                faye: {
+                    subscription: existingSubscription
+                }
+            }));
+
+            const promise = mockedStore.dispatch(fayeActions.subscribe());
+            promise.should.equal(existingSubscription);
+
+            return promise
+                .then(() => {
+                    getClientStub.should.not.have.been.called;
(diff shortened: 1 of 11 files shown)


test/specs/actions/faye.spec.js
```diff
@@ -342,11 +342,76 @@ describe('Faye Actions', () => {
...
-            mockedStore = createMockedStore(sandbox, generateBaseStoreProps());
-            mockedStore.dispatch(fayeActions.subscribe()).then(() => {
-                setFayeSubscriptionSpy.should.have.been.calledOnce;
+        let getClientStub;
+        let subscribeStub;
+
+        beforeEach(() => {
+            subscribeStub = sandbox.stub().resolves();
+
+            getClientStub = sandbox.stub().returnsSyncThunk({
```
alavers: I think if we didn't have `returnsSyncThunk` this code could be:

```
cosnt client = {
    subscribe: subscribeStub
}
getClientStub = sandbox.stub().returns(() => client);
```

which is a lot clearer if you ask me /2cents

test/specs/actions/integrations.spec.js
```diff
@@ -24,62 +30,171 @@ describe('Integrations Actions', () => {
...
-                httpStub.should.have.been.calledWith('POST', `/apps/${appId}/appusers/${_id}/clients`, {
-                    criteria: {
-                        type: 'twilio',
-                        phoneNumber: '+0123456789'
-                    },
-                    confirmation: {
-                        type: 'prompt'
+        describe('when the user exists', () => {
+            beforeEach(() => {
+                httpStub.returnsAsyncThunk({
```
alavers: I think I have a similar comment here

```
httpStub.returns(() => Promise.resolve({client: createdClient}))
```

test/specs/components/Settings.spec.jsx
```diff
@@ -94,17 +85,4 @@ describe('Settings', () => {
         const description = TestUtils.scryRenderedDOMComponentsWithClass(component, 'settings-description')[0];
         description.textContent.should.eq(mockedStore.getState().ui.text.notificationSettingsChannelsDescription);
     });
-
-    describe('No user id', () => {
```
alavers: Hmm, surprised this simply gets removed Should this test expectation simply be changed to `should render` ?

mspensieri: I think the rest of the tests are covering that case

test/specs/actions/integrations.spec.js
mspensieri: Marc was doing that a while ago, and the first time I read it I was like "why wouldn't you just use `stub.resolves()`?", then I noticed that it's actually returning a _function_ which returns a promise when invoked - it wasn't immediately obvious what the function was doing there. I think `returnsAsyncThunk` hides some of that confusion, even if it brings confusion of it's own.

Both styles are kind of ugly, but I'd rather keep `returnsAsyncThunk` and `returnsSyncThunk` since that's what's used in all the other tests

test/specs/components/Settings.spec.jsx
alavers: Oic. Store props doesn't have a user by default any more in the `beforeEach`

test/specs/actions/integrations.spec.js
alavers: If you guys both find the thunk style clearer I'll go along 👍

test/specs/actions/faye.spec.js
lemieux: I'd like to rewrite the utils to be like `returnsThunk().val(val)`  for sync and `returnsThunk().resolves(val)` or `returnsThunk().rejects(val)` for async..... bu that's for another day...

PR #589, opened 2017-08-16
Fix link events with user switching

mspensieri:
While addressing code review comments from #585 and #587 I realized
that link events were handled incorrectly when the user changed. The
`login` event now expects a `userId` and `jwt` as params, which we
weren't passing. Also, we were still connected to the the faye channel
for the old discarded user.

### DIFF
--- a/src/frame/js/actions/faye.js
+++ b/src/frame/js/actions/faye.js
@@ -3,8 +3,8 @@ import { batchActions } from 'redux-batched-actions';
 
 import { setUser } from './user';
 import { showSettings, showTypingIndicator, hideTypingIndicator, hideChannelPage, hideConnectNotification } from './app-state';
-import { addMessage, incrementUnreadCount, resetUnreadCount, getMessages, fetchUserConversation } from './conversation';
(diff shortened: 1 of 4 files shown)


src/frame/js/actions/faye.js
```diff
@@ -161,10 +163,11 @@ function updateUser(currentAppUser, nextAppUser, client) {
                 setItem(`${appId}.sessionToken`, nextAppUser.sessionToken);
             }
 
+            dispatch(disconnectFaye());
```
wmora: would it also work if `disconnectFaye()` is included in the actions array passed to `batchActions(actions)` instead?

mspensieri: No because `disconnectFaye` is a thunk and `batchActions` expects raw action objects

lemieux: no because batchActions doesn't support thunks... I know it's confusing.

wmora: ok thanks

alavers: I asked the exact same question the other day...

PR #591, opened 2017-08-16
Remove appUserId from storage if the user has a userId

lemieux:
Pretty much what the title and the comment in the code say.


lemieux:
I'm adding tests for `handleUserConversationResponse`

PR #593, opened 2017-08-17
Use promise rejections instead of throwing

lemieux:
While I was preparing the testing phrase, I realize how the expectations for the public function were screwed up. If you were calling `init` with bad props, it would throw. If props were good but auth was bad (bad appId for instance), it would return a rejected promise. 

I made everything return a rejected promise if that function was meant to return a promise when it's working alright. The current version works like that anyway.

### DIFF
--- a/test/specs/smooch.spec.js
+++ b/test/specs/smooch.spec.js
@@ -283,13 +283,13 @@ describe('Smooch', () => {
                         });
 
                         it('should reject', () => {
-                            Smooch.init(props).should.be.rejectedWith(Error, /already initialized/);
+                            return Smooch.init(props).should.be.rejectedWith(Error, /already initialized/);
                         });
                     });
 
                     describe('without appId', () => {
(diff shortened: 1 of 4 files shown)


test/specs/smooch.spec.js
```diff
@@ -282,16 +282,14 @@ describe('Smooch', () => {
                             }));
                         });
 
-                        it('should throw', () => {
-                            (function() {
-                                Smooch.init(props);
-                            }).should.throw(/already initialized/);
+                        it('should reject', () => {
+                            Smooch.init(props).should.be.rejectedWith(Error, /already initialized/);
```
mspensieri: I think you'll need to `return` here, `rejectWith` should return a promise which mocha should wait on. Same with the tests below, `return` or `yield`

lemieux: Done... yield is not a thing in the frontend.

PR #595, opened 2017-08-21
Make fake promises chainable

lemieux:
The fake promises returned by the loader and by the host lib weren't chainable. Now they are.

### DIFF
--- a/src/frame/js/web-messenger.jsx
+++ b/src/frame/js/web-messenger.jsx
@@ -215,6 +215,7 @@ export function init(props = {}) {
         .catch((err) => {
             Raven.captureException(err);
             cleanUp();
+            throw err;
         });
 }
 
--- a/src/frame/js/web-messenger.jsx
+++ b/src/frame/js/web-messenger.jsx
@@ -194,13 +194,13 @@ export function init(props = {}) {
                                     storage.removeItem(`${props.appId}.appUserId`);
                                     storage.removeItem(`${props.appId}.sessionToken`);
 
-                                    return store.dispatch(batchActions([
+                                    store.dispatch(batchActions([
                                         authActions.resetAuth(),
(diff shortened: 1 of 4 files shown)


test/specs/web-messenger.spec.js
```diff
@@ -203,7 +203,7 @@ describe('WebMessenger', () => {
                                     appId: hat()
                                 };
 
-                                return WebMessenger.init(props).then(() => {
+                                return WebMessenger.init(props).catch(() => {
```
mspensieri: nit: `.init(props).should.be.rejected.then(() => { })` to make sure that a successful init doesn't pass this test (and the others below)

lemieux: good point

src/frame/js/web-messenger.jsx
```diff
@@ -194,13 +194,13 @@ export function init(props = {}) {
                                     storage.removeItem(`${props.appId}.appUserId`);
                                     storage.removeItem(`${props.appId}.sessionToken`);
 
-                                    return store.dispatch(batchActions([
+                                    store.dispatch(batchActions([
```
mspensieri: 😕  why change this logic? This is a case where we recovered from bad `appUserId`/`sessionToken` and just booted as a fresh anonymous user instead. I think the init can be considered a success here

lemieux: hmm well then the test assertion was bad

mspensieri: I think it was just accidentally changed here https://github.com/smooch/smooch-js/pull/595/commits/d87409c5ca6980c97a82785c1aec200b849b839a#diff-980e1568ab82d9ffbf8e1c57ff157d00L206

/shrug

PR #597, opened 2017-08-23
Fix image upload

lemieux:
Image upload wasn't working for multiple reasons : 1. bad data sent, 2. all message data was appended to the form when it wasn't needed.

This should fix it once the backed is also fixed.

### DIFF
--- a/src/frame/js/actions/conversation.js
+++ b/src/frame/js/actions/conversation.js
@@ -113,11 +113,10 @@ function postSendMessage(message) {
 
 function postUploadImage(message) {
     return (dispatch, getState) => {
-        const {config: {appId}, user: {_id}} = getState();        
+        const {config: {appId}, user: {_id}} = getState();
         const blob = getBlobFromDataUrl(message.mediaUrl);
         const data = new FormData();
 
-        data.append('source', blob);
         data.append('sender', JSON.stringify({
             type: 'appUser',
             client: getClientInfo(appId)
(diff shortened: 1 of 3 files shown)


src/frame/js/actions/conversation.js
```diff
@@ -112,19 +113,19 @@ function postSendMessage(message) {
 
 function postUploadImage(message) {
     return (dispatch, getState) => {
-        const {config: {appId}, user: {_id}} = getState();
+        const {config: {appId}, user: {_id}} = getState();        
         const blob = getBlobFromDataUrl(message.mediaUrl);
-
         const data = new FormData();
+
         data.append('source', blob);
```
mspensieri: Can we put `source` as the last object to be appended? It gives some flexibility to optimize validation server-side later on

src/frame/js/index.js
```diff
@@ -1,4 +1,4 @@
-import './utils/polyfills';
+// import './utils/polyfills';
```
mspensieri: ?

src/frame/js/actions/conversation.js
```diff
@@ -112,19 +113,19 @@ function postSendMessage(message) {
 
 function postUploadImage(message) {
     return (dispatch, getState) => {
-        const {config: {appId}, user: {_id}} = getState();
+        const {config: {appId}, user: {_id}} = getState();
```
mspensieri: format gone wrong

src/frame/js/index.js
lemieux: ugh, leftover of a test

src/frame/js/actions/conversation.js
lemieux: sure thing

src/frame/js/actions/conversation.js
lemieux: waa

PR #598, opened 2017-08-24
Throw error if app is inactive

lemieux:
An inactive app shouldn't be allowed to render. This fixes it.

### DIFF
--- a/src/frame/js/web-messenger.jsx
+++ b/src/frame/js/web-messenger.jsx
@@ -179,7 +179,7 @@ export function init(props = {}) {
             const {config: {app: {status}}} = store.getState();
 
             if (status !== 'active') {
-                return Promise.resolve();
+                throw new Error('App is inactive.');
             }
 
             return Promise.resolve()
--- a/src/frame/js/web-messenger.jsx
+++ b/src/frame/js/web-messenger.jsx
@@ -99,6 +99,13 @@ function cleanUp() {
     store.dispatch(reset());
 }
 
+function checkInitialization() {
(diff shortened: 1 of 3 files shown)


src/frame/js/web-messenger.jsx
```diff
@@ -179,7 +179,7 @@ export function init(props = {}) {
             const {config: {app: {status}}} = store.getState();
 
             if (status !== 'active') {
-                return Promise.resolve();
+                throw new Error('App is inactive.');
```
mspensieri: The call to `cleanUp` on line 218 fixes the case where you can still send a message to an inactive app, right?

lemieux: Yeah it should reset the store state. I'm maybe thinking about adding a `isInitialized` check on each public method to be cleaner. Currently, it would let you call them and fail with `Unauthorized`.

test/specs/web-messenger.spec.js
```diff
@@ -425,19 +526,64 @@ describe('WebMessenger', () => {
...
+                mockedStore = mockAppStore(sandbox, generateBaseStoreProps({
+                    ...defaultState,
+                    appState: {
+                        isInitialized: false
+                    }
+                }));
+            });
+
+            it('should throw an error', () => {
+                return expect(() => WebMessenger.open()).to.throw(Error, /initialize the Web Messenger first/);
```
mspensieri: `.close` here

lemieux: :man_facepalming:

PR #601, opened 2017-08-24
Add upgrade flow for non-enterprise sdk users

mspensieri:
Added logic to upgrade non-enterprise sdk users into enterprise sdk
users by calling the upgrade user endpoint. To detect if an upgrade is
necessary, we check for a legacy client id stored in the browser
storage. If an id is found, we call the upgrade endpoint to
retrieve the user.

If the call is successful, we use the returned appUserId and
sessionToken to continue the initialization flow. If the call is
unsuccessful, we continue as a new anonymous user. In either case, the
clientId is moved out of legacy storage and into the new storage,
meaning this API will not be called again the next time the page is
loaded.


src/frame/js/actions/auth.js
```diff
@@ -81,6 +81,18 @@ export function logout() {
     };
 }
 
+export function upgradeUser(clientId) {
+    return (dispatch, getState) => {
+        const {config: {appId}} = getState();
+        return dispatch(http('POST', `/apps/${appId}/appusers/upgrade`, {
+            clientId
+        }))
+            .then((response) => {
```
lemieux: nit : you could one line this

test/specs/utils/client.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    describe('getClientId', () => {
+        it('should return legacy client id', () => {
+            const legacyId = hat();
+            getItemStub.withArgs('sk_deviceid').returns(legacyId);
+
+            const id = clientUtils.getClientId(appId);
+
+            id.should.eql(legacyId);
+
+            getItemStub.should.have.been.calledOnce;
```
alavers: There are two different styles of this that are a bit more pretty/concise:

```
getItemStub.should.have.been.calledOnce.calledWith('sk_deviceid');
```

or, if you don't care about non-matching call count:

```
getItemStub.withArgs('sk_deviceid').should.have.been.calledOnce;
```

But in the end to each their own :D

PR #604, opened 2017-08-28
Fix faye implementation in git fork

jugarrit:
This fixes a problem in the faye lib we're using where it was failing to correctly fall back to long polling if websockets did not work.

See https://github.com/faye/faye/pull/482 + https://github.com/faye/faye/pull/483 for details on the fixes.

The https://github.com/jugarrit/faye/tree/temp-smooch-fix branch contains both fixes and will be used in the meantime.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "license": "SEE LICENSE IN LICENSE",
   "dependencies": {
     "deep-equal": "1.0.1",
-    "faye": "1.1.2",
+    "faye": "jugarrit/faye#temp-smooch-fix",
     "ismobilejs": "0.3.9",
     "lodash.bindall": "4.4.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -148,8 +148,5 @@
     "babel-runtime": ">=6.9.0 <7.0",
     "react": "15.x.x",
     "react-dom": "15.x.x"
-  },
-  "browser": {
-    "faye": "faye/browser/faye-browser"
```
mspensieri: Was this unused?

lemieux: Pretty sure it was useless, it's not even in `f/compat-expansion`.

jugarrit: It might have been used in the older version of Faye, but the new one has the "browser" directive in its package.json for sure.

PR #609, opened 2017-08-30
Fix bug causing SDK to make multiple failed connection attempts

jugarrit:
Changing the faye branch to one that is better maintained...

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "license": "SEE LICENSE IN LICENSE",
   "dependencies": {
     "deep-equal": "1.0.1",
-    "faye": "smooch/faye#faye-fixes",
+    "faye": "smooch/faye#fa283a9d074d9add7dee173f897b8d691ec6f6f5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
   "license": "SEE LICENSE IN LICENSE",
   "dependencies": {
     "deep-equal": "1.0.1",
-    "faye": "smooch/faye#temp-smooch-fix",
+    "faye": "smooch/faye#faye-fixes",
```
lemieux: maybe it would be easier to target the commit sha instead of creating branches each time..?

jugarrit: My thought was that from now on we'd use this branch. I wanted a place where I could open PRs to if we needed further fixes.

lemieux: Might be easier to trigger new deploys (and see what actually changed) if we change the sha each time though

jugarrit: Sure

PR #612, opened 2017-08-30
Update README to reflect new behavior

lemieux:
I also fixed the loader generation task as it was compressing the script too much.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,18 +18,21 @@ The Smooch Web Messenger will add [live web messaging](https://smooch.io/live-we
 
 ### Script Tag
 
-Add the following code towards the end of the `body` section on your page. Placing it at the end allows the rest of the page to load first.
-
-```html
-<script src="https://cdn.smooch.io/smooch.min.js"></script>
-```
-
-
-Initialize the plugin using this code snippet
+Add the following code towards the end of the `head` section on your page and replace `<app-id>` with your app id at the end of the script.
 
 ```html
 <script>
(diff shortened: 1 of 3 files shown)


jugarrit:
Is there supposed to be some readme changes here?

lemieux:
:man_facepalming:

README.md
```diff
@@ -18,18 +18,21 @@ The Smooch Web Messenger will add [live web messaging](https://smooch.io/live-we
 
 ### Script Tag
 
-Add the following code towards the end of the `body` section on your page. Placing it at the end allows the rest of the page to load first.
+Add the following code towards the end of the `head` section on your page and replace `<app-id>` with your app id at the end of the script.
 
 ```html
-<script src="https://cdn.smooch.io/smooch.min.js"></script>
+<script>
```
jugarrit: Oooh the script got longer.
Why does the appId need to be provided on `init` if it's already provided in this script?

lemieux: Because the loader and `init` are two different things. `init` is the same whether you use it from our cdn or from npm. We could have something where it's cached in memory and `init` would use that, but I didn't feel it was clean enough to have different ways of calling `init` dependending on how you install Smooch.

alavers: Also I think this loader will be a templated string you can just copy paste from the smooch dashboard. If we go that route I wonder if these instructions should just be a link to those dashboard instructions.

lemieux: Yup, we should update this part once it's live on the webapp

PR #640, opened 2017-10-02
Upload props on login, return pending props with getUser()

lemieux:
This PR fixes two things : 
- Pending props were not handled properly when a user was logged in and there were pending props to be uploaded
- `getUser()` was not returning the pending props if there was any and it gave the impression the props were lost in the void.

I ended up creating a new reducer to hold pending props. There was a `pendingAttributes` key in the user props, but I search for the place where it's actually updated, and turns out, it didn't happen. Pending props were previously stored in a variable inside the User action scope which was kind of bad.

### DIFF
--- a/src/frame/js/actions/conversation.js
+++ b/src/frame/js/actions/conversation.js
@@ -9,6 +9,7 @@ import { disconnectClient, subscribe as subscribeFaye, unsetFayeSubscription } f
 import http from './http';
 import { setAuth } from './auth';
 import { setConfig } from './config';
+import { resetPendingUserProps } from './user';
 
 import { observable } from '../utils/events';
 import { Throttle } from '../utils/throttle';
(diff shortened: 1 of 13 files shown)


src/frame/js/actions/conversation.js
```diff
@@ -577,10 +578,11 @@ export function startConversation() {
                 });
         } else {
             promise = dispatch(http('POST', `/apps/${appId}/appusers`, {
-                ...pendingAttributes,
+                ...pendingUserProps,
                 userId,
                 client: getClientInfo(appId)
             }));
+            dispatch(resetPendingUserProps());
```
alavers: Should `resetPendingUserProps` wait for the POST to finish?

lemieux: hmm, right, otherwise the `getUser()` behavior would be a bit weird during that time.

